PR #18, opened 2020-08-04
initial multiArch support

### LINKED ISSUE #17: Multiarch support?
Not necessarily, although i don't think this will work with arm builds. With the growing rise of arm as a desktop/laptop replacement, it's important to have templates like these adaptable to those needs.

I don't have any solution to provide, but I do see that there are some x86-64 assumptions in `flake.nix` and probably other places.  Ideally down to generating an image for an arm system would work. For example, have `hosts/niximg-x86-64.nix` and `hosts/niximg-aarch64.nix`

**Additional context**
I have been monitoring and studying this repository for a few weeks now and am definitely interested and like how things are shaping up.  I've been leaning towards using it in place of what ive already started, but I have several arm-based machines that I want to work well with my configuration.

### LINKED ISSUE #34: nix-darwin support
Not really. Would like to have `nix-darwin` support as well.

Understand that this is currently targeted towards NixOS, but would be great to have support for [`nix-darwin`](https://github.com/LnL7/nix-darwin#flakes-experimental) builds as well.

**Additional context**
Unfortunately, I am just getting started with `flakes` (and generally `nix`), but think the philosophy of having a repository to get up and running on `nix` could apply across NixOS and macOS.

nrdxp:
This should resolve #17 and fix #34. NixOS is only supported on x86 Linux, so this does not affect nixosConfigurations. This only affects the `devShell` and `package` outputs. NixOS is pinned to x86_64 for now.

Every package is tested to ensure it is compatible with a given system based on its platform metadata _before_ it it added to that systems package set.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,20 @@
 {
   "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1605370193,
+        "narHash": "sha256-YyMTf3URDL/otKdKgtoMChu4vfVL3vCMkRqpGifhUn0=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "5021eac20303a61fafe17224c087f5519baed54d",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
     "home": {
       "inputs": {
         "nixpkgs": "nixpkgs"
@@ -78,6 +93,7 @@
     },
     "root": {
       "inputs": {
+        "flake-utils": "flake-utils",
         "home": "home",
         "master": "master",
         "nixos": "nixos",
--- a/flake.nix
+++ b/flake.nix
@@ -6,15 +6,17 @@
       master.url = "nixpkgs/master";
       nixos.url = "nixpkgs/release-20.09";
       home.url = "github:nix-community/home-manager/release-20.09";
+      flake-utils.url = "github:numtide/flake-utils";
     };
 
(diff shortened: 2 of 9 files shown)


tgunnoe:
Hello! This is quite the turnaround from what I assumed to be a dead request.  Thanks!
> NixOS is only supported on x86 Linux, so this does not affect nixosConfigurations. This only affects the `devShell` and `package` outputs. NixOS is pinned to x86_64 for now.

What do you mean by "NixOS is only supported on x86 Linux"? do you mean officially? I haven't seen any statements about official support, but nixpkgs does support multiarch, but maybe not as much non-x86 builds. For example I do still have a hard time updating from both channels `20.03` and `unstable` on my Pinebook Pro. I think it's mostly the newness of `aarch64` using `nixpkgs`, like at some point i heard the builds are failing in Hydra but were recently fixed.

I'm referring to you for my understanding of `flakes` using a system configuration.


> NixOS is pinned to x86_64 for now.

Thus I do not understand what this means, as nixpkgs has no defined arch? I'm still relatively new and only started tracking [my own NixOS configs](https://github.com/tgunnoe/configurations) only recently, but this simple repo layout has been working fine in between `x86` and `aarch64` machines so far.

nrdxp:
@tgunnoe, sorry for the delayed response on this. I made the assumption, based on the images available for download on nixos.org, that NixOS was only support on x86, since those are the only images available. It is possible that it would work on other architectures.

Although, since the flakes specification doesn't allow for defining `nixosConfigurations.<some-system>` without breaking, I'd assume the answer is no. I think I'm finally ready to implement this since `flake-utils` would make it rather trivial. I just sort of forgot about it with all the crazy stuff going on this year.

PR #49, opened 2020-12-29
Add an `apps` output attribute set

nrdxp:
Apps are automatically derived from the `packages` output. Only packages that contain an "$out/bin" are packaged as apps.


blaggacao:
Couldn't this (and the filtering) even go upstream into flake-utils?

nrdxp:
I had a similar thought. Another potential canidate for upstream would be a template in NixOS/templates.

PR #52, opened 2020-12-29
ditch ls for exa

blaggacao:


blaggacao:
just saw it's already there

nrdxp:
I have similar aliases defined in develop/zsh, along with a zsh function to autopage if the output doesn't fit on one line. The function is zsh dependant, but I don't want the aliases to clash. Here is how the are defined. I originally kept exa out of core but it's light enough that it should be fine. Here are my aliases.

```nix
{
      ls = "exa";
      l = "ls -lhg --git";
      la = "l -a";
      t = "l -T";
      ta = "la -T";
}
```

Looks like it's just `l` `la` and `ls` that clash.

We could move the zsh aliases to `programs.zsh.shellAliases` or yours to `programs.bash.shellAliases` instead of the global `environment.shellAliases` to avoid collision if we want to keep both.

PR #61, opened 2020-12-30
ref: remove overlay xref to pkgs

blaggacao:
since packages are exposed through output.packages by their own,
a second export through overlays.pkgs would be smudging the public api
of this repository.

### DIFF
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -85,11 +85,10 @@ in
   genPackages = { overlay, overlays, pkgs }:
     let
       packages = overlay pkgs pkgs;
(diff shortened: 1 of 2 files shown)


blaggacao:
I see that `pkgs` are implemented as overlays, as well. Consequentially, I see an opportunity to re-export local packages simply by [namespacing them](https://github.com/nrdxp/nixflk/issues/60), as in [this example](https://github.com/numtide/flake-utils/blob/08c7ad4a0844adc4a7f9f5bb3beae482e789afa4/simpleFlake.nix#L52)

nrdxp (approved):
Your right, there is no reason to have pkgs in both `overlays.pkgs` and as the `overlay`.

PR #62, opened 2020-12-30
update home-manager urls

lourkeur:
HM moved to nix-community on Github. There is a 301 redirect, but I have no idea how long it will stay up, so it's better not to rely on it.

Things done:
- [x] Grep for rycee
- [x] Click both links on Github to make sure that they work
- [x] Test configuration with `NIXOS_CONFIG=$PWD/configuration.nix nixos-option`
- [x] Run `nixpkgs-fmt --check configuration.nix`


nrdxp:
Thank you for the PR. However the url was already updated in #25. Sounds like you are working off of an old revision, possibly.

PR #64, opened 2020-12-30
Bugfix: flk shell command always printing help.

Maj-majormajormajor:
There is a bug in the first conditional of the shell script. It always evaluates the conditional `"--help"` as true. Fix: check against first argument.


Maj-majormajormajor:
Fixed by https://github.com/nrdxp/nixflk/commit/6d96a5888fe7135485a45b97e566758f66bcf427

Issue #58, opened 2020-12-30
use github actions for cachix ci


nrdxp:
I've been, so far, pushing builds to my cachix manually, but an automated github action would allow every package to be built, on every branch, including the devshells and inputs.

A [reference](https://nix.dev/tutorials/continuous-integration-github-actions.html) implementation will have to be adapted to this project, but it shouldn't be too difficult.

nrdxp:
duplicate of #59. Got a 404 when submitting so didn't realize it actually go through the api.

PR #71, opened 2021-01-10
Add nrdxp cachix to substituter flake list

blaggacao:
courtesy of: https://github.com/NixOS/nix/commit/343239fc8a1993f707a990c2cd54a41f1fa3de99

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,4 +1,6 @@
 {
+  nixConfig.substituters = [ "https://nrdxp.cachix.org" ];
+
   description = "A highly structured configuration database.";
 
   inputs =


nrdxp:
What is the benefit of this over the current cachix configuration?

blaggacao:
> What is the benefit of this over the current cachix configuration?

Between the two api choices (clasical & nflake), I'd argue there might be a slight bias towards choosing the flake-native api.

No materialized bottom-line benefit other than "api consistency" (of some sort).

blaggacao:
> No materialized bottom-line benefit other than "api consistency" (of some sort).

I'm currently not so sure about this, any more. after adding a couple of substituters (including the official nix cache and yours) in my flake, it seems that `flk iso` doesn't attempt to build `glibc` (and similar low level stuff) any more...

nrdxp:
If I understand this correctly, it looks like this allows the usage of binary cache for any user, not just those who have it setup in their system.

blaggacao:
> If I understand this correctly, it looks like this allows the usage of binary cache for any user, not just those who have it setup in their system.

Correct, it is kind enough to ask:

```console
do you want to allow configuration setting 'substituters' to be set to 'https://cache.nixos.org/ https://nrdxp.cachix.org https://thefloweringash-armv7.cachix.org' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? N
```

But there is a caveat:
https://github.com/numtide/flake-utils/pull/18/files#diff-79c5531512c8fa0e4522e5d82d96fc2d93b75040e0a197cb0a94c7cb0b2c19b3R96

nrdxp:
We'll probably want to leave this open til the fate of that PR is decided.

nrdxp:
I'm gonna go ahead and merge this now that #58 is resolved, it'll be much more useful.

blaggacao:
I figured this does not actually seem work in my current nix version. Don't know what's happening.

Trying to share a local store, like so
- `nix --option substituters "http://myip:5000" develop`
I also had to do
- `nix --option substituters "http://myip:5000" --options require-sig false develop`
for what it's worth.

PR #75, opened 2021-01-10
Instruct reloading nix shell after changes

blaggacao:
At least in my case, adding https://github.com/nrdxp/nixflk/pull/74/files required leaving and re-entering (`nix develop`) the nix shell for the override to take effect.

Maybe there is a better way to alert & inform the user. It seems to be always necesary to reload.

Or alternatively: maybe best solution would be to do the required re-instantiation directly within `flk` command, since re-entering the nix shell is a bit awkward UX

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,9 @@ a template. There is a [bare branch][bare] if you want to start with a
 completely empty template and make your own profiles from scratch. The only
 hard requirement is nix itself. The `shell.nix` will pull in everything else.
 
+When using nix shell for building your system generation, after a change you
+need to exit and re-enter the nix-shell.
+
 ## Flake Templates
 If you already have [nix-command][nix-command] setup you can:
 ```sh


nrdxp:
I believe I've run into this a few times already. At least for direnv setups, we could easily solve this with something like `watch_file **/*.nix`, so that the shell reinitializes after any nix file changes.

In the general case though, I'm not sure what the best solution is. I didn't notice this problem until we switched to using mkDevShell, so I'm wondering if that is somehow the source of the issue.

blaggacao:
I could imagine the "culprit" is https://github.com/nrdxp/nixflk/blob/e67cd3701aa2a2081d0c261ccc12b47f743c0866/shell.nix#L10

where `./.` might be the store path representation of the source tree.

```
  configs = builtins.trace "${toString ./.}#nixosConfigurations" "${toString ./.}#nixosConfigurations";
```
yields
```
trace: /nix/store/55y5c9rj4fiv7jvxr9mpvk6prlyg8rl2-source#nixosConfigurations
```

Which doesn't get updated until upon rebuild.

Not sure what to do here. Maybe we could use `DEVSHELL_ROOT` env variable here?

blaggacao:
Superseded by #87

PR #77, opened 2021-01-11
Evaluate exported pkgs against repo baseline nixos

### LINKED ISSUE #76: ovoverlay import mechanics don't seem to allow file name attribute discrepancy and also do not distinguish between nixos and master
Because of
https://github.com/nrdxp/nixflk/blob/d0fe508da20695bd72c1a5eebd3cb113778366ac/flake.nix#L51

in conjunction with

https://github.com/nrdxp/nixflk/blob/d0fe508da20695bd72c1a5eebd3cb113778366ac/lib/utils.nix#L89-L92

https://github.com/nrdxp/nixflk/blob/d0fe508da20695bd72c1a5eebd3cb113778366ac/overlays/kakoune.nix#L13

would not be made available in `pkgs`.

In general, I'm not sure if "too much magic" here is really a win or if overlays could be managed in a more transparent manner.

I really like the way how flake-utils's simpleFlake commands name spacing (see #60).
That approach doesn't limit overlaying "stock" packages, though, through a little extra helper that would read overlays from a non  name spaced portion of the overlay file:

```nix
final: prev: {
   nixflk = {}; # name spaced overlays -- sharing friendly as `self.overlay` (namespaced, too!)
   utillinux = prev.util-linux; # non name spaced overlays -- ought not to be  exported as `self.overlay` (sic!) but available under `self.overlays.utillinux`
}
```

---

On the other hand it seems that we need a way to target an overlay either at `master` or at `nixos`, since on the contrary we expose the users to [nasty and unexpected side effects](https://github.com/nrdxp/nixflk/issues/63#issuecomment-757566201) of the implementation detail that `genPackages` targets both branches.

blaggacao:
The baseline pkgs for own consumption is nixos, only pkgs/override.nix
provides a mechanism to "backport" packages from unstable.

Therefore, we should expose our packages to others in the same
context as we use them ourselves.

---

I was about to open an issue, yet I figured it would be more helpful to directly propose a PR, instead. Maybe this needs more preliminary discussion, though.

See also #76 (part about `master` vs `nixos` overlays)

---

Proporsal together with #78 

fix #76

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,7 +64,7 @@
         let
           pkgs = pkgImport {
             inherit system;
-            pkgs = master;
+            pkgs = nixos;
             overlays = [ devshell.overlay ];
           };
 


nrdxp:
Probably a good idea. Won't merge until we fully decide on a solution to #76.

PR #78, opened 2021-01-11
Don't evaluate overlays on master

### LINKED ISSUE #76: ovoverlay import mechanics don't seem to allow file name attribute discrepancy and also do not distinguish between nixos and master

blaggacao:
master is only used in the context of `pkgs/override.nix`
and `unstableModulesPath`
and only within `./hosts`

Since [overlays exists that are incompatible between nixos and master
at evaluation time (sic!)](https://github.com/nrdxp/nixflk/issues/63#issuecomment-757566201), this together with #77 avoids evaluating
overlays against master all together.

---

Proposal to

fix #76

### DIFF
--- a/lib/utils.nix
+++ b/lib/utils.nix
@@ -41,7 +41,8 @@ in
       };
 
       unstablePkgs = pkgImport {
-        inherit system overlays;
+        inherit system;
+        overlays = [];
         pkgs = master;
       };
     };


blaggacao:
Goes a little into the direction / assumptions of: https://github.com/numtide/flake-utils/pull/18/files#diff-79c5531512c8fa0e4522e5d82d96fc2d93b75040e0a197cb0a94c7cb0b2c19b3R126-R132

nrdxp:
I'm gonna have to test this, as I think it might break my qutebrowser overlay.

blaggacao:
I was thinking, if an overlay for a master package is needed, could this be conceptualized as a `./pkgs` overlay, perhaps? After all those are a fair amount of time the ones one would strive to upstream anyway.

It would thinks easier, and wouldn't incommode too much?

nrdxp:
This doesn't seem to break anything. I'll go ahead an merge. I still want a better interface to pulling unstable overrides, but we can deal with that in a separate issue/pr.

PR #80, opened 2021-01-11
Little hint

blaggacao:


flake.nix
```diff
@@ -3,6 +3,7 @@
 
   inputs =
     {
+      # nix flake update --update-input master
```
bew, line 6: Can you add a one line description of what the command does?
Otherwise it looks like a copy/paste typo if you know know what is does. (I don't)

blaggacao: True

blaggacao: However, we can't put the circustances of "desired" in here. That wouldn't be terse enough and also shift quickly as the ecosystem evolves...

PR #81, opened 2021-01-11
shell: add `flk up` command

nrdxp:
Automatically generate a configuration for your current system using nixos-generate-config, placed in the `up` directory. A file is added in hosts/up-$(hostname).nix importing said configuration.

Useful for getting up and running quickly, but if the user intends to use more than one host, they should probably move the generated bits directly into their host config manually.
 
@blaggacao, would appreciate your input.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 result
+up
+hosts/up-*
 .direnv
--- a/README.md
+++ b/README.md
@@ -122,6 +122,15 @@ flk install yourConfig # deploys hosts/yourConfig.nix
 flk yourConfig switch
 ```
 
+### Note on `flk up`:
+While the `up` sub-command is provided as a convenience to quickly set up and
+install a "fresh" NixOS system on current hardware, committing these files is
+discouraged.
+
+They are placed in the git staging area automatically because they would be
+invisible to the flake otherwise, but it is best to move what you need from
+them directly into your hosts file and commit that instead.
+
 ## Build an ISO
 
 You can make an ISO and customize it by modifying the [niximg](./hosts/niximg.nix)
--- a/README.md
+++ b/README.md
@@ -104,12 +104,9 @@ for ease of use. Once inside:
 # This will setup nix-command and pull in the needed tools
 nix-shell # or `direnv allow` if you prefer
 
-# quick way to setup your fileSystems (assuming they are partioned):
-sudo mount /dev/<install-drive> /mnt
-nixos-generate-config --root /mnt --show-hardware-config > hosts/yourConfig.nix
-
-# Edit your config to add a bootloader
-$EDITOR hosts/yourConfig.nix
+# use nixos-generate-config to generate a basic config for your system
+# edit hosts/up-$(hostname).nix to modify.
+flk up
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,2 +1,4 @@
 result
+up
+hosts/up-*
```
blaggacao, line 3: What motivates the exclusion of `hosts/up-*` from VC?

nrdxp: every configuration generated by `flk up` is prefixed with "up-". This actually might not end up being a good idea after all though, since files have to be staged in git for them to be visible in the flake.

blaggacao: I faintly wondered :-)
We could use this: https://github.com/NixOS/nixpkgs/blob/f0a7fce6ee72e574c2406bf1cfd1741c8d2cd07e/nixos/modules/installer/tools/nixos-generate-config.pl#L589-L590

blaggacao: but it leaves out some aspects... :-(

nrdxp: I was considering it, but since the bootloader is set in `configuration.nix`, it's somewhat incomplete and requires new users to manually declare a bootloader. The point of the `up` command is take care of this for new users so they can get up and running as quickly as possible.

blaggacao: It seems as if the only way then is to namespace the folder per hostname,:

`./hw-auto-gen/hostname/` ?

blaggacao: I like `up` better, but maybe a 2 line readme would have it's merrit declaring that all files are autogoenerated and should not be edited directly or somthing...

blaggacao: `echo "Files in this folder are created by flk up command - do not edit directly" > ./up/AUTO-GENERATED` ... just dropping in some thoughts.

nrdxp: I originally wanted to discourage users from commiting the autogenerated bits without integrating them properly into their hosts, however, since git requires staging, namespacing might be the better solution after all.

nrdxp:
Added a commit which namespaces `flk up` and stages the files, since the flake can't find them without it.

.gitignore
blaggacao, line 3:
```suggestion

```

blaggacao:
Looks pretty useful already!

.gitignore
nrdxp: I think we should still ignore these by default and force add them as I've done here, no?

blaggacao: Oh, I overlooked the `-f` :smile: .

Thinking about the effects, not sure if it's worth it:

Once added to the index, it doesn't prevent the user from modifying them and committing on the same file.

What is the intended effect under the new situation (that flakes requires it to be in the index)?

nrdxp: Really, the intention is more psychological than technical. Adding these to gitignore signifies that we really don't want to be commiting stuff from these directories, but instead should take what we need from them and add it elsewhere. We could reinforce this by adding something to the README

blaggacao: >We could reinforce this by adding something to the README

I think, for proper message conveyance, that's good idea.

PR #82, opened 2021-01-11
Hosts: fix mod override

blaggacao:
In some occasions the module path was renamed.
To avoid conflicts, the old path must be disabled manually.

E.g.
```nix
{
  unstableModules = [
    "services/ttys/getty.nix"
  ];
  addToDisabledModules = [
    "services/ttys/agetty.nix"
  ];
}
```


blaggacao:
This is only an escape hatch, no prime api surface.

PR #85, opened 2021-01-11
settings: initial toml configuration

nrdxp:
Create a `settings.toml` file for easily setting configuration wide dependant options. Includes the example of setting the interactive user via this file.

Profiles requires introspective knowledge of the user configuration which is not normally available. The example, implemented here, is the interactive system user. Some profiles need to add this user to a specific group. 

Previously we relied on the users profile to keep track of, and add these users to the required groups. Now, a profile can pull the currently declared username from `settings.users.interactive` and add that user to the required groups. This improves profiles ability to provide a hermetic configuration for specific applications, while simultaneously simplifying configuration for new users.

Pull in settings from any NixOS module like so:
```nix
{ settings, ... }:
{
  users.users."${settings.users.interactive}" = {
    # ...
  };
}
```
Definitely want some feedback before committing to a final spec.
@blaggacao @jtojnar @bbigras @lourkeur @codygman @adamscott

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -24,8 +24,8 @@ let
 
       specialArgs =
         {
+          inherit settings;
           unstableModulesPath = "${master}/nixos/modules";
-          settings = fromTOML (readFile ../settings.toml);
         };
 
       modules =
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -11,7 +11,7 @@
 }:
 let
   inherit (utils) recImport;
-  inherit (builtins) attrValues removeAttrs;
+  inherit (builtins) attrValues fromTOML readFile removeAttrs;
   inherit (pkgset) osPkgs unstablePkgs;
 
   unstableModules = [ ];
@@ -24,6 +24,7 @@ let
       specialArgs =
         {
           unstableModulesPath = "${master}/nixos/modules";
+          settings = fromTOML (readFile ../settings.toml);
         };
 
       modules =
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -4,7 +4,7 @@
 , master
 , pkgset
 , self
-, system
+, settings
 , utils
 , externModules
 , ...
@@ -13,6 +13,7 @@ let
   inherit (utils) recImport;
   inherit (builtins) attrValues fromTOML readFile removeAttrs;
   inherit (pkgset) osPkgs unstablePkgs;
+  inherit (settings) system;
 
   unstableModules = [ ];
   addToDisabledModules = [ ];
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -15,8 +15,8 @@ let
   inherit (builtins) attrValues fromTOML readFile removeAttrs;
   inherit (settings) system;
 
-  unstableModules = [ ];
-  addToDisabledModules = [ ];
+  unstableModules = settings.modules.unstable;
+  addToDisabledModules = settings.modules.disabled;
 
   config = hostName:
     lib.nixosSystem {
@@ -31,6 +31,16 @@ let
       modules =
         let
           core = self.nixosModules.profiles.core;
+
+          profiles =
+            let
+              inherit (settings.meta) machine machines;
+            in
+            {
+              imports = map
+                (profile: self.nixosModules.profiles.${profile})
+                machines.${machine} or [ ];
+            };
 
           modOverrides = { config, unstableModulesPath, ... }: {
             disabledModules = unstableModules ++ addToDisabledModules;
@@ -76,6 +86,7 @@ let
           core
           global
           local
+          profiles
           modOverrides
         ] ++ externModules;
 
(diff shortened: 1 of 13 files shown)


jtojnar:
Not really sure what this is trying to achieve but it feels to me like it is trying to do two or three things at the same time:

- setting global option – for this we can already import a custom NixOS module that defines custom options, with the added benefit that we get schema validation
- specifying configuration values in a TOML file – this might be useful for some file this might be useful for some people but I still do not see a reason for parallel option infrastructure
- using unstructured/schema-less/ad-hoc options – this can be achieved in NixOS modules as well, and @infinisil’s work makes it even better with partial settings schemas

nrdxp:
The only motivation for toml is to simplify configuration for new users, perhaps those not familiar or comfortable with nix. I was considering maybe just adding some settings to the `lib` attribute as an alternative.

blaggacao:
> The only motivation for toml is to simplify configuration for new users

I think this is inline with this project's goals. Similar to how `flk` is. So I think it adds value. We need to keep monitoring evolution of nickel in light of such accommodations, though.

nrdxp:
We probably want to add the value of `system` into the toml as well, so new users have a single place to modify defaults before initial deployment. Essentially, we want to discourage users from digging into the guts of the flake, and make it obvious where configuration is supposed to happen.

blaggacao:
> Essentially, we want to discourage users from digging into the guts of the flake, and make it obvious where configuration is supposed to happen.

I'd add that we can also encourage users peeking at [a clean `flake.nix`](https://github.com/numtide/flake-utils/pull/18/files#diff-7c6d831f44328c9c097362771f76c4d43bd055eb169865821449b920ae2bb0ef), provided we can reduce its complexity. That doesn't preclude a `settings.toml`, though.

hosts/default.nix
```diff
@@ -24,6 +25,7 @@ let
       specialArgs =
         {
           unstableModulesPath = "${master}/nixos/modules";
+          settings = fromTOML (readFile ../settings.toml);
```
blaggacao, line 28: I'd prefer `nixflkSettings` to be a good citizen and properly "own" things by name spacing them (and preserve public faith: it's not nixos native)

profiles/core/default.nix
```diff
@@ -1,13 +1,17 @@
-{ config, lib, pkgs, ... }:
-let inherit (lib) fileContents;
-
+{ config, lib, pkgs, settings, ... }:
+let
+  inherit (lib) fileContents;
+  inherit (settings) users;
```
blaggacao, line 4: side note: as a newcomer I find this technique to use global name spaces a bit harder to read.

blaggacao:
`backportPkgsFromAlt` (master) would also be an interesting candidate...

`./pkgs/overrides.nix` is pretty plain for it to be buried in the folder hierarchy.

blaggacao:
Similarily, maybe the `list.nix` values could be consolidated. It makes reasoning a lot easier if such things which serve an indexing purpose to the repository are not spread over the folder hierarchy.

blaggacao:
&rarr; "The magic ought to be the sole powers of `settings.toml`"

hosts/default.nix
blaggacao: Better: `nixflk.settings` (!)

blaggacao:
Re/ `overrides.nix`

Sring package names can be resoled like this: https://github.com/numtide/flake-utils/pull/18/files#diff-79c5531512c8fa0e4522e5d82d96fc2d93b75040e0a197cb0a94c7cb0b2c19b3R70-R76

(taken from `numtide/devshell`)

blaggacao:
`settings.toml` &rarr; `nixflk.toml` (again: good citizenship)

nrdxp:
Thanks for the good suggests. `settings` is definitely a place holder, `nixflk` sounds a lot better. Consolidating the lists makes a lot of sense as well.

I went ahead and did the obvious thing for the package overrides f2aedb889b770beff161a89d959b262e62d4d2df.  Not sure if it will now map simply to toml.

blaggacao:
I traced https://github.com/nrdxp/nixflk/commit/f2aedb889b770beff161a89d959b262e62d4d2df in https://github.com/numtide/flake-utils/pull/18/commits/414c8790081d035c4226d034f9e408087e8de976.

nrdxp:
After using this for a while, I'm not convinced it is the best approach after all, however I do want to introduce some mechanism to abstract collections of profiles into reusable chunks independant of any given system. Perhaps dhall whould be a better target here than toml?

blaggacao:
Is there no way to do it in toml (even if that means a little less DRY)? Maybe those collections can be defined in `nix`?

The collection idea is interesting: I have a restaurant project where I might want to try using collections, indeed: "waitress" / "cashier"

nrdxp:
Well, I've already done it in TOML, if you look at the settings.toml you'll see a `meta.machines` which can be a bunch of different named collections, and the user sets one with `meta.machine`. Perhaps even a simple NixOS module would be nice, so users could simply:
```nix
{
  settings.machines.myMachine = [
    # list of profiles
  ];

  settings.machine = "myMachine";
}
```

PR #87, opened 2021-01-12
Fix realod

blaggacao:
better fix for #75

---

had no chance to test, yet. But am confident.


nrdxp:
It works, and appears to solve the problem. Thanks.

PR #89, opened 2021-01-12
shell: fix pass overlays to devshell as well

blaggacao:
see also: https://github.com/numtide/flake-utils/pull/18/files#diff-79c5531512c8fa0e4522e5d82d96fc2d93b75040e0a197cb0a94c7cb0b2c19b3R178-R187

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -68,7 +68,7 @@
             pkgs = pkgImport {
               inherit system;
               pkgs = nixos;
-              overlays = [ devshell.overlay ];
+              overlays = externOverlays ++ [ self.overlay ] ++ (builtins.attrValues self.overlays);
             };
 
             packages = flattenTreeSystem system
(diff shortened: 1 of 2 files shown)


blaggacao:
Apparently it yet doesn't do what I intended.

blaggacao:
I'm not sure if we generally want to do this since in casu it only affects nixflk's own shell.nix. The reason I wanted to attempt it is to pass along `nixUnstable`, so that we have better control over latest and greatest nix features. (I'm not sure how much stable branch lags behind, yesterday it was just a few days f a release).

While dogfooding the overlays is probably right, I'm not entirely sure about the overrides.

On the other hand, thinking twice, maybe passing overlays to specifically `nixflk/shell.nix` maybe is overkill and should be avoided.

So I'll retract here and adapt my flake-utils PR accordingly with a hint for reasoning.

blaggacao:
I correct: we need it upstream in flake-utils for a potentially more generic use case, but we don't need it here for the current `nixflk/shell.nix`. If flake-utils should be adapted, we could get it for free, then.

PR #93, opened 2021-01-14
users: Initial home-manager only configurations(non-nixos systems)

### LINKED ISSUE #83: Home manager on Non-nixos systems
I believe that at the moment any home manager configuration you describe in nixflk cannot be used on a non-nixos system, even though this is a known use case of home manager.

An easy way to apply home manager configurations on non-nixos systems with the existing solution.

 This could maybe be some command that takes in a user and applies the home-manager.users.<input user> to the current home directory.

 Or perhaps each user in home-manager.users could get an export in the flake.nix so they can be applied with nixos-rebuild(I think this is possible). I'm still learning flakes, so I'm not sure about implementation details. 

I think home manager configurations can be exported as packages with the activationPackage attribute, so maybe a package for each user.

**Describe alternatives you've considered**
I think you could make a home.nix that takes in the relevant config and pass that to hm switch.
Also there might already be a way to do this with nixflk and I just don't know.

**Additional context**
This was discussed on the home-manager repo: https://github.com/nix-community/home-manager/issues/1510. It looks like the configuration input for home-manager is a file, so I'm not sure how that would work with the way nixflk is formatted, since the users files include other things for the nixos configuration.

Pacman99:
Closes #83 

Working implementation of exporting home-manager only configurations. This exports to a non-official output of homeConfigurations, along with hmActivationPackages in the packages output. Both are an attrset of hosts then users

So if you were trying to build activationPackage for the host NixOS and user nixos, you can do:
`nix build ./#homeConfigurations.NixOS.nixos.home.activationPackage`
or 
`nix-build ./#hmActivation.NixOS.nixos'
    This ones easier for debugging, because nix build will default to the packages output if homeConfiguration has any error and it won't say when it does so.

The flk script can then use the activation package to activate the home-manager configuration(not implemented yet).

There are a few more things to add like local nix registry and config with xdg.configFile along with nix paths and other compatibility things.

TODO:
 - [x] Create homeConfiguration for each user in each system
 - [x] Export activation packages for each homeConfiguration
 - [x] Add support in flk wrapper script to easily build and switch home-manager configurations
 - [x] Document feature in readme
 <s>- [ ] Add some default modules/imports to all the homeConfigurations, preferably only in homeConfigurations and not for system configs. Not sure if this is even possible yet. This can include: nix paths, local nix.conf, and others in the future.</s>I think this can be done in the future, possibly along with adding custom modules feature to home-manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,43 @@ They are placed in the git staging area automatically because they would be
 invisible to the flake otherwise, but it is best to move what you need from
 them directly into your hosts file and commit that instead.
 
+## Home Manager Integration
+The home-manager nixos module is available for each host. It is meant
+to be used in the user profiles, you can find an example in the nixos user profile
+
+The home-manager configuration for each user in each system is available in the
+outputs as homeConfigurations and the activation packages in hmActivationPackages.
+
+This allows you to just build the home-manager environment without the rest of the
+system configuration. The feature is useful on systems without nixos or root access.
+
+Lets say you want to activate the home configuration for the user `nixos` in the 
+host `NixOS`.
+
+With the flk script:
+```sh
+# You can build it using
+flk home NixOS nixos
+# and activate with
+./result/activate
+
+# Or do both like this
+flk home NixOS nixos switch
+```
+
+This can also be done manually:
+```sh
+
+# With hmActivationPackages, what the flk script uses
+nix build ./#hmActivationPackages.NixOS.nixos
+
+# Or with homeConfigurations, 
+nix build ./#homeConfigurations.NixOS.nixos.home.activationPackage
+# this is hard to debug though, due to nix build's fallback to packages
+
+# The configuration can then be activated like before
+```
+
 ## Build an ISO
 
 You can make an ISO and customize it by modifying the [niximg](./hosts/niximg.nix)
--- a/flake.nix
+++ b/flake.nix
@@ -29,7 +29,8 @@
       inherit (builtins) attrValues;
       inherit (flake-utils.lib) eachDefaultSystem flattenTreeSystem;
       inherit (nixos.lib) recursiveUpdate;
-      inherit (self.lib) overlays nixosModules genPackages pkgImport;
+      inherit (self.lib) overlays nixosModules genPackages pkgImport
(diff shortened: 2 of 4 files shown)


nrdxp:
I haven't had a lot of free time this month, but I'll be glad to test this out as soon as I've got a minute. 

Shouldn't be too difficult to work out the kinks.

Pacman99:
Yeah of course take your time, thanks for responding. I'll try to keep working through the bugs and implementing a few other things. I think overlays aren't currently added to the pkgs passed into home-manager and the flake registry and nix path can be set through home manager. 

I also have one question, I put the code in default.nix in the users folder, but it directly uses nixosConfigurations, so it doesn't actually use any of the user files. Does it make sense to have it there? I was thinking it could go in utils.

Pacman99:
So this turned out to be a lot simpler than I thought, apparently you can just use config.home-manager.users.[username].home.activationPackage to build the hm configuration. 

I also appended the homeConfiguration activationPackages to the packages output as hmActivationPackages so you can just build and debug with ./#hmActivationPackages.[host].[user]

I think this is a pretty much complete and fairly simple implementation of this. There are a couple things left to do and I'll add that to the main post.

nrdxp:
Just pushed a rebase PR to your branch to fix the merge issues.

Pacman99:
Alright thanks for doing the rebase! I think I merged it right, I just force pushed you're exact branch to mine

nrdxp:
Only minor issue left when running `nix flake check`:
```
flake attribute 'packages.x86_64-linux.hmActivationPackages' is not a derivation
```

I think we can resolve this later though as it doesn't actually affect functionality.

Pacman99:
Ohh interesting. How does it work in nixpkgs, since that has subsets of packages like pythonPackages? Or is pythonPackages a derivation itself somehow?

nrdxp:
Right now, nixpkgs doesn't even use the `packages` output attribute it it's flake, instead cramming everything into `legacyPackages` for now. Pressumably because of issues like this that still need to be resolved upstream.

Pacman99:
Ohh that makes sense

PR #96, opened 2021-01-19
nixos-hardware: use the flake instead of a path

benneti:
Just after you merged my PR I realised there is also a more an approach to include nixos-hardware more inline with the idea of falkes.
This basically means no paths are need to include the hardware configuration and looks a bit cleaner in my opinion.


nrdxp:
Agreed! Thank you.

PR #103, opened 2021-01-27
shell: use devshell-native pre-commit hooks

blaggacao:
Since https://github.com/numtide/devshell/pull/63, git hooks where introduced into deshell.

They better encapsulate the development environment than pre-commit does. Notably, they are under nix control not under pre-commit control.

@zimbatm looks good, first adoption attempt in the wild :+1: Thx!


nrdxp:
This is beautiful, thank you!

Issue #72, opened 2021-01-10
ARM aarch64 Support


blaggacao:
Quite a few people might want to get up to speed with deploying their RPi (eg. Home automation, etc).

I'm going to think of a way on how to integrate it into flk so that the UX of getting started with raspberry (and other board) builds is cool.

blaggacao:
I'm realizing that for RPi builds, it would make much more sense if the sd base configuration would actually be a "hidden" (from `flk`) module and instead each host would gain a `flk sd <host>` command. The same thing might actually be true for `iso` as well...

In other words, people would be encouraged to do `flk iso <host>` so to base off their ISO on a specific host config.

nrdxp:
This could potentially be out of scope for this project, since we are primarily concerned with providing a template for NixOS personal use.

blaggacao:
Alright, closing ... :+1: 

Maybe the `flk iso <host>` would be a good abstraction though (and be very sd flashing compatible)?

blaggacao:
I'm trying to play the patient game and get [this merged](https://github.com/NixOS/nixpkgs/pull/110827), so we can have (in theory) a SD image build.

Of course that would not do anything about package compatibility and such...

tgunnoe:
similar to https://github.com/divnix/devos/issues/17

blaggacao:
indeed, I think the missing piece are cross builds, as far as I can tell. I'm slowly getting to the point where I need crossbuilds to work, so I'll be probably following through on this. If you have any pointers or ideas, please let me know.

I want to be able to do `flk sd NixOS` so that NixOS will be crossbuild (in aarch64) for flashing on a sd card.

PR #105, opened 2021-01-28
host: add netboot default build target

blaggacao:
Same as #102 but for netboot (includes final version of #102)


blaggacao:
@nrdxp I adapted this to the iso build target. So you would just have to decide wether to support the netboot target. Use cases can be if somebody wants to manage RPIs in home automation and instead of switching sd cards take out this step by netbooting (and such).

nrdxp:
Please rebase this pr on core, and target it to merge there as per #112

nrdxp:
We will still include this in a fix for #72

PR #109, opened 2021-02-01
profiles: add concept of suites

nrdxp:
An alternative to #85. I wanted a mechanism for defining arbitrary users and profiles without mixing in concerned for specific hardware so we can easily pass them around and eventually generate different kinds of machines from them, e.g. containers.

- [x] Import attrs of profiles automatically with `defaultImport`.
- [x] Refactor profiles to ensure all are functions returning a module.
- [x] Add a suites.nix with collections of profiles.
- [x] Add suites as `specialArgs` to modules.
- [x] Add suite import to NixOS host.
- [x] same for users


nrdxp:
@blaggacao, any comments?

blaggacao:
The naming is spot on. 👍

PR #116, opened 2021-02-08
core: fix #115

### LINKED ISSUE #115: process information is hidden by default
as the profile also contains `hideProcessInformation = true;` and this one interferes with some display managers at the moment, it probably should be disabled, too.
(see https://github.com/NixOS/nixpkgs/issues/111629)

_Originally posted by @benneti in https://github.com/nrdxp/nixflk/issues/114#issuecomment-774919998_

Similarly to #114, a hardening option is included in `core`. In addition to benneti's comment, I wish to point out that this renders non-root gotop (as well as other `*top`s) unable to display useful information as well.

# Expected behavior
gdm+wayland and `*top` programs should Just Work as long as the user doesn't opt-in to hardening settings.

lourkeur:
As discussed in related issues, this shouldn't be on by default and is likely a leftover from the earlier stages of the project.

### DIFF
--- a/profiles/core/default.nix
+++ b/profiles/core/default.nix
@@ -141,8 +141,6 @@ in
     '';
   };
 
-  security.hideProcessInformation = true;
-
   services.earlyoom.enable = true;
 
   users.mutableUsers = false;


nrdxp:
Is this also contributing to the hibernation issue?

lourkeur:
I don't believe it is. When I went to fix hibernation on my setup I didn't need to disable that.

nrdxp:
Either way, I think this is still a good idea. I like hiding proc info personally, but by default is a bit aggressive.

PR #135, opened 2021-02-23
pkgs: automatic source management

### LINKED ISSUE #118: semi automatic update for /pkgs
I do not want to check manually whether my custom pkgs have updates.

Add https://github.com/ryantm/nixpkgs-update to the configuration and maybe a litte subcommand to flk to have a simple interface to it (I plan to try it myself when I have some more time).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96566401-semi-automatic-update-for-pkgs?utm_campaign=plugin&utm_content=tracker%2F161955272&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F161955272&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nrdxp:
Fixes #118.

- [x] Leverage flakes to manage package sources & hashes
- [x] Update documentation with an example.
- [x] Add `mkVersion` function to autogenerate a version string
- [x] Add srcs package via overlay containing all sources defined in  _pkgs/flake.nix_
- [x] Extend `flk update` with the ability to only update the given input
- [x] Fix flake-compat support 

Would appreciate any feedback before finalizing. 

My only nitpick is that I'd love to add the ability for `mkVersion` to pull a flake urls ref if one is given. For example, a flake url "github:owner/repo/version-tag" would generate the version: "version-tag". Doesn't seem like flake inputs keep the original ref though, only the sha, so we can't do this without parsing the lock file, or potentially filing an issue upstream to include this information.

> ##### _Note:_
> ~~This will break hercules-ci, since it currently relies on flake-compat, and this technique [breaks](https://discourse.nixos.org/t/possibility-for-subflakes/11589/6) flake-compat. So I'm kinda hesitant to merge unless a workaround is found.~~

/cc @benneti  @blaggacao @Pacman99

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,10 @@
 # Introduction
 DevOS grants a simple way to use, deploy and manage [NixOS][nixos] systems for
 personal and productive use. It does this by providing a convenient repository
-structure; integrating several popular projects like
-[home-manager][home-manager], and [devshell][devshell].
+structure, integrating several popular projects like
+[home-manager][home-manager], and [devshell][devshell], and offering useful
+conveniences like
+[automatic source updates](./pkgs#automatic-source-updates).
 
 Skip the indeterminate nature of other systems, _and_ the perceived difficulty
 of Nix. It's easier than you think!
--- a/doc/flk/update.md
+++ b/doc/flk/update.md
@@ -4,3 +4,13 @@ The `update` subcommand is a simple alias for:
 nix flake update --recreate-lock-file --commit-lock-file
 ```
 As it sounds, this will update your lock file, and commit it.
+
+## Updating Package Sources
+If you pass an input name then it will only update that input.
+
+For example, you can update any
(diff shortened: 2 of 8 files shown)


blaggacao:
> Doesn't seem like flake inputs keep the original ref though, only the sha, so we can't do this without parsing the lock file, or potentially filing an issue upstream to include this information.

https://github.com/NixOS/nix/issues/3838

I was never sold on @edolstra 's counter argument, but this might count as one interesting use case.

blaggacao:
Works perfect except for `vendorSha256` shenanigans. I was never a fan of that construct and wonder why I'm supposed not to trust the chain of git commit -> go.sum lockfile shas.

PR #144, opened 2021-03-01
pkgs/flake: add comments to retrieve new versions

blaggacao:
I find such comments generally handy and helpful when the time to update comes.

### DIFF
--- a/pkgs/flake.nix
+++ b/pkgs/flake.nix
@@ -2,18 +2,23 @@
   description = "Package sources";
 
   inputs = {
+    # git ls-remote https://github.com/libretro/retroarch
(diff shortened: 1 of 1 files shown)


nrdxp:
Perhaps the [docs](https://devos.divnix.com/pkgs/index.html#automatic-source-updates) would be a better spot for this. In general, I want to get rid of all comments and move as much as possible to the docs. 

I may end up moving forward with the dox project I showed you, as nixdoc is generating xml, which we really can't use, and it hasn't been updated in 2 years. Once that's done, we can use doc comments for library functions, but most everything else should live in the docs.

I also would love to extend mdbook at some point with a backend for pulling and evaluating nix code, so we can ensure the examples in the docs actually work.

PR #145, opened 2021-03-01
lib: export namespaced lib as normal overlay

blaggacao:
##### Offspring: #147 

---

todo: a flollowup commit could change caller sites within this repo to
use `pkgs.devos.lib` / `nixos.devos.lib`, as well.

---

(Hypothetical) lib-only usage
```nix
{
  description = "Flake utils demo";

  inputs.flake-utils.url = "github:numtide/flake-utils";
  inputs.devos.url = "github:divnix/devos";

  outputs = { self, nixpkgs, flake-utils, mach-nix }:
    flake-utils.lib.simpleFlake {
      inherit self nixpkgs;
      name = "my-flake";
      preOverlays = [ devos.overlay ];
    };
}
```

and then:

```nix
{ pkgs, ...}:
   pkgs.devos.lib."[...]"
```

---

Namespaceing with litteral `devos` implies sticky ownership of those parts by `divnix/devos`.

I reason that this might be the right way of thinking about the `./lib` folder, though:

- If you make changes, chances are good you wanted to upstream them anyway at some point (or maintain a private fork of `devos.lib`)
- I'm not sure if I would like to maintain repo-specific library functions in `./lib` since that makes syncing with the mothership ever more difficult. I've started by placing repo-specific library functions under `./pkgs`

On the other hand, we might also conclude that it is better to namespace with the current project's name, regardless if those are upstreamable changes and encourage to put all repo-specific library functions under `./lib`.

Thinking out loud, I'm actually not sure any more. Maybe upgrading `./lib` is the better road of the two.

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -64,20 +64,12 @@ let
 
                 nixos = {
                   exact = true;
-                  from = nodes.nixos.original;
+                  from = {
+                    id = "nixos";
+                    type = "indirect";
+                  };
                   to = {
                     inherit (nixos) lastModified narHash rev;
-
-                    path = override.outPath;
-                    type = "path";
-                  };
-                };
-
-                override = {
-                  exact = true;
-                  from = nodes.override.original;
-                  to = {
-                    inherit (override) lastModified narHash rev;
 
                     path = override.outPath;
                     type = "path";


blaggacao:
This is sort of the model adopted by https://github.com/numtide/flake-utils/tree/master/examples/simple-flake. Sure for back-compat, we could leave the `lib` export.

If this is a good idea or not depends on whether `lib` will become a first class flake citizen different from `packages`.

If that is the case, there ought to be contrived a solution to also pass `lib` when doing overlays.

blaggacao:
Well, in the latter commit, I did it: dynamic name spacing :stuck_out_tongue_winking_eye:

PR #147, opened 2021-03-01
lib: can depend on pkgs (a la nixpkgs#pkgs/pkgs-lib)

blaggacao:
A tiny step carved out of thought journey while working on #145.

Benefit: we can encourage/document to extend `./lib` more usefully:

example:
```nix
{ nixos, pkgs, ... }:
# [...]
{
  # [...]
  # Custom Libaray Utilities
  mkPlutoApp = pkgs.callPackage ./mkPlutoApp { };
}
```

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -8,6 +8,7 @@
   - [Cachix](./cachix/README.md)
   - [Extern](./extern/README.md)
   - [Hosts](./hosts/README.md)
+  - [Lib](./lib/README.md)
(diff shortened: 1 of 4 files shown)


nrdxp:
looks interesting, don't have enough time to test right now. Will checkout later and look around.

nrdxp:
This looks fairly benign, and since it's mostly documentation, which is always great, I'll go ahead and merge it. Thank you!

PR #148, opened 2021-03-01
modules: document service extension

blaggacao:
It took me a while to figure that out, so I thought it would be worthwhile documenting for others.

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -87,7 +87,6 @@ modules/services/cluster/k3s/extended.nix
 
 let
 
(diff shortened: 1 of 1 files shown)


nrdxp:
I'm okay with this, but have you ensured it actually works? From what I understood from NixOS/nixpkgs#16884, extending and/or overriding modules was not possible.

blaggacao:
I havn't yet validated this. Should have been draft and I just found an error 🤷‍♂️ .

PR #155, opened 2021-03-09
Expose namespaced lib on pkgs

blaggacao:
I'm not 100% sure if the local library is already supercharged on `lib`, but also for sharing as an overlay I think a name spaced identifier is a better solution in any case.

Currently, I adopted `myLib`, but that should probably become `myname.lib` so that I can use library functions more cleanly from fellow devossers. Though, I didn't want to propose that together with this PR, since it is an independent change / discussion.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,7 +63,7 @@
               inherit (pkgs) lib;
             });
 
-          overlay = import ./pkgs;
+          overlay = import ./pkgs { inherit nixos; };
 
           lib = import ./lib { inherit nixos pkgs; };
 
--- a/lib/README.md
+++ b/lib/README.md
@@ -82,6 +82,15 @@ lib/default.nix:
 }
 ```
 
+You can then access this library function via `pkgs.myLib`:
+
+```nix
+{ pkgs, ... }:
+
+pkgs.myLib.mkCustomI3BindSym { "..." };
+
+```
+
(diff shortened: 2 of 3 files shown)


nrdxp:
I'm not clear on the benefit here. Can you perhaps give a good example?

blaggacao:
> I'm not clear on the benefit here. Can you perhaps give a good example?

Hm, ok. I'll try to explain myself better:

- I am assuming that `pkgs.lib` does not carry the custom library functions of `.lib` (I might be erred on that already).
- Then I write a custom library function that I want to use in my environment.
- Then I want to use that custom function in `./pkgs` or `./modules`
- My best answer to "How do I get the custom function there?" was "Expose it on `pkgs`", but using `pkgs.lib` would have been tinkering with the global name space, which I saw as risky and prone of eventual future naming conflicts. Hence, I figured `pkgs.myLib` would solve this partially. Partially only, because if everyone name spaces with `pkgs.myLib` and we start sharing library functions, then we end up with name conflicts again. A better solution would have been `pkgs.baggacao.lib` but this would have been too much of a change for this PR as it requires passing the user name arroud and I think needs proper evaluation separately.

nrdxp:
Are you aware that lib is already available via `lib.flk`? After #157 is merged, it will be `lib.dev`.

blaggacao:
Ah ok, I was already mislead on the first point. 👍

Issue #125, opened 2021-02-17
host for aarch64 system


liaowang11:
**Describe the bug**
Adding an aarch64 system with aarch64 specific packages (like libraspberrypi) gives error:
```
Package ‘libraspberrypi’ in /nix/store/widqcq445r5r4ivh7sdfibhjrpq69jgl-source/pkgs/development/libraries/libraspberrypi/default.nix
:19 is not supported on ‘x86_64-linux’, refusing to evaluate.
```
**To Reproduce**
1. Add `raspi4.nix` to `/hosts`
2. run `flk raspi4 dry-run`
3. errors with content like above

```nix
#raspi4.nix
#…
  environment.systemPackages = with pkgs; [                                     
      libraspberrypi                                                              
   ]; 
#…
```
**Expected behavior**
Dry run should succeed

**Additional context**
I’m trying to use flake to config all my machines. This template is the best I can find, but still does not fully support multi-arch. Have tried hlissner’s dotfiles, also not working.
The dotfiles from [Dunklecat](https://git.sr.ht/~dunklecat/nixos-config/tree/master/item/flake.nix) is the only one I ‘ve found support both aarch64 and x86_64. Hope `flk` can support it.

nrdxp:
The only reason I haven't wired up multiarch for configs is because the flake spec doesn't expect system attributes under `nixosConfigurations`, and adding them screws up nixos rebuild and friends. 

I have been thinking about the best way to resolve this, and perhaps a non standard output is the best way for now. You can of course, change the system:
https://github.com/nrdxp/nixflk/blob/77296ff433c0dc6e9b5cad522b6a46dafaa204ba/flake.nix#L53

And it should work, but this will change the system for all your hosts. I think I'll just add a `nixos` output with aarch and i686 configs later today.

Pacman99:
AIUI the reason theres no multi-arch nixosConfigurations output like the packages output works, is that it doesn't make sense for the same nixosConfiguration to be used under multiple archs. If you want a aarch64 nixosConfiguration output, your supposed make the nixosSystem specifically for that aarch64 system. And that makes sense with nixflk's structure since each host is meant for one system, so you just make a new host file for your aarch64 system.

[...]

PR #161, opened 2021-03-14
Multi-arch support for hosts(nixosConfigurations)

### LINKED ISSUE #72: ARM aarch64 Support

Pacman99:
fixes #72 

also related, #125 

This allows users to set `nixpkgs.system` to any architecture exported by the nixpkgs flake and nixpkgs.pkgs will be set to that system.

I also added `multiPkgs` as a special arg and made `nixpkgs.pkgs` a default. So if someone wanted to do crossSystem builds or anything else with pkgs, that is also an option(eg. mobile-nixos!!).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,47 +36,45 @@
 
       extern = import ./extern { inherit inputs; };
 
-      pkgs' = os.mkPkgs;
+      multiPkgs = os.mkPkgs;
 
-      outputs =
-        let
-          system = "x86_64-linux";
-          pkgs = pkgs'.${system};
-        in
-        {
-          nixosConfigurations =
-            import ./hosts (nixos.lib.recursiveUpdate inputs {
-              inherit pkgs system extern;
-              inherit (pkgs) lib;
+      outputs = {
+        nixosConfigurations =
+          import ./hosts (nixos.lib.recursiveUpdate inputs {
+            inherit multiPkgs extern;
+            defaultSystem = "x86_64-linux";
+            lib = nixos.lib.extend (final: prev: {
+              dev = self.lib;
             });
+          });
 
-          nixosModules =
-            let moduleList = import ./modules/module-list.nix;
-            in lib.pathsToImportedAttrs moduleList;
+        nixosModules =
+          let moduleList = import ./modules/module-list.nix;
+          in lib.pathsToImportedAttrs moduleList;
 
-          overlay = import ./pkgs;
-          overlays = lib.pathsToImportedAttrs (lib.pathsIn ./overlays);
+        overlay = import ./pkgs;
+        overlays = lib.pathsToImportedAttrs (lib.pathsIn ./overlays);
 
-          lib = import ./lib { inherit nixos pkgs self inputs; };
+        lib = import ./lib { inherit nixos self inputs; };
 
-          templates.flk.path = ./.;
-          templates.flk.description = "flk template";
-          defaultTemplate = self.templates.flk;
+        templates.flk.path = ./.;
+        templates.flk.description = "flk template";
+        defaultTemplate = self.templates.flk;
 
-          deploy.nodes = os.mkNodes deploy self.nixosConfigurations;
+        deploy.nodes = os.mkNodes deploy self.nixosConfigurations;
 
-          checks =
-            let
-              tests = import ./tests { inherit self pkgs; };
-              deployChecks = builtins.mapAttrs
-                (system: deployLib: deployLib.deployChecks self.deploy)
-                deploy.lib;
-            in
-            nixos.lib.recursiveUpdate tests deployChecks;
-        };
+        checks =
+          let
+            tests = import ./tests { inherit self pkgs; };
+            deployChecks = builtins.mapAttrs
+              (system: deployLib: deployLib.deployChecks self.deploy)
+              deploy.lib;
+          in
+          nixos.lib.recursiveUpdate tests deployChecks;
+      };
 
       systemOutputs = utils.lib.eachDefaultSystem (system:
-        let pkgs = pkgs'.${system}; in
+        let pkgs = multiPkgs.${system}; in
         {
           packages = utils.lib.flattenTreeSystem system
             (os.mkPackages { inherit pkgs; });
--- a/flake.nix
+++ b/flake.nix
@@ -62,20 +62,21 @@
         defaultTemplate = self.templates.flk;
 
         deploy.nodes = os.mkNodes deploy self.nixosConfigurations;
-
-        checks =
-          let
-            tests = import ./tests { inherit self pkgs; };
-            deployChecks = builtins.mapAttrs
-              (system: deployLib: deployLib.deployChecks self.deploy)
-              deploy.lib;
-          in
-          nixos.lib.recursiveUpdate tests deployChecks;
       };
 
       systemOutputs = utils.lib.eachDefaultSystem (system:
         let pkgs = multiPkgs.${system}; in
         {
+          checks =
+            let
+              tests = nixos.lib.optionalAttrs (system == "x86_64-linux")
+                (import ./tests { inherit self pkgs; });
+              deployHosts = nixos.lib.filterAttrs
+                (n: _: self.nixosConfigurations.${n}.config.nixpkgs.system == system) self.deploy.nodes;
+              deployChecks = deploy.lib.${system}.deployChecks { nodes = deployHosts; };
+            in
+            nixos.lib.recursiveUpdate tests deployChecks;
+
           packages = utils.lib.flattenTreeSystem system
             (os.mkPackages { inherit pkgs; });
 
--- a/hosts/README.md
+++ b/hosts/README.md
@@ -25,6 +25,10 @@ that you intend to use on your machine.
 Additionally, this is the perfect place to import anything you might need from
 the [nixos-hardware][nixos-hardware] repository.
 
+> ##### _Note:_
+> Set `nixpkgs.system` to the architecture of this host, default is "x86_64-linux". 
+> Keep in mind that not all packages are available for all architectures.
+
 ## Example
 
 hosts/librem.nix:
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -3,9 +3,9 @@
 , lib
 , nixos
 , override
-, pkgs
+, multiPkgs
 , self
-, system
+, defaultSystem
 , ...
 }:
 let
@@ -27,7 +27,7 @@ let
           modules;
       };
 
-    global = {
+    global = { config, ... }: {
       home-manager.useGlobalPkgs = true;
       home-manager.useUserPackages = true;
 
@@ -39,7 +39,7 @@ let
         "home-manager=${home}"
       ];
 
-      nixpkgs = { inherit pkgs; };
+      nixpkgs.pkgs = lib.mkDefault multiPkgs.${config.nixpkgs.system};
 
       nix.registry = {
         devos.flake = self;
@@ -79,7 +79,8 @@ let
       };
     in
     dev.os.devosSystem {
-      inherit system specialArgs;
+      inherit specialArgs;
+      system = defaultSystem;
       modules = modules // { inherit local lib; };
     };
 
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -1,4 +1,4 @@
-args@{ nixos, pkgs, self, ... }:
+args@{ nixos, self, ... }:
 let inherit (nixos) lib; in
 lib.makeExtensible (final:
   let callLibs = file: import file
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -59,45 +59,44 @@ let
     nixosTesting.makeTest calledTest;
 in
 {
-  x86_64-linux = {
-    profilesTest = mkTest {
-      name = "profiles";
+  profilesTest = mkTest {
+    name = "profiles";
 
-      machine = { suites, ... }: {
-        imports = suites.allProfiles ++ suites.allUsers;
-      };
-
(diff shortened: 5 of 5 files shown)


hosts/NixOS-a64.nix
```diff
@@ -0,0 +1,14 @@
+{ suites, ... }:
+{
+  ### root password is empty by default ###
+  imports = suites.base;
+
+  nixpkgs.system = "aarch64-linux";
```
Pacman99, line 6: This entire file is probably unnecessary and I could just add that line to the hosts README. But it is nice for testing the PR.

lib/default.nix
```diff
@@ -1,4 +1,4 @@
-args@{ nixos, pkgs, ... }:
```
Pacman99, line 1: I don't think lib functions should ever need to depend on pkgs. I think nixos.lib should be enough.

hosts/default.nix
```diff
@@ -62,7 +63,9 @@ let
       modOverrides
     ] ++ extern.modules;
 
-  specialArgs = extern.specialArgs // { inherit suites; };
+  specialArgs = extern.specialArgs // {
+    inherit suites multiPkgs;
```
nrdxp, line 67: Do we need `multiPkgs` as a specialArg? When would one system ever need access to anothers package set? Are you thinking for cross compiling?

lib/default.nix
nrdxp: agreed, @blaggacao, you added this. Could you explain why it was added and if it's okay to remove from your point of view? Also, I believe the [docs on lib](https://devos.divnix.com/lib/index.html) need to be updated regardless.

nrdxp (changes requested):

hosts/default.nix
Pacman99: Yeah exactly, I was specifically thinking cross compiling. Since you can no longer set localSystem and crossSystem in a devos host. But I'm not sure if its needed either, idk much about cross compiling.

hosts/NixOS-a64.nix
nrdxp: Instead of this file, I would write a NixOS integration test similar to the [profilesTest](https://github.com/divnix/devos/blob/core/tests/default.nix#L63-L73)

hosts/default.nix
nrdxp: So one can already build cross packages for a given platform from the `pkgsCross` meta package. However, maybe this could still be useful. 

One should be able to build these `multiPkgs` with the proper binfmt registered though. For example, on the hercules-ci runner i have set: `boot.binfmt.emulatedSystems = [ "aarch64-linux" ]` so it can build and test the aarch packages. This might be a good idea after all so users can cross-compile their overlay or user defined packages.

Pacman99: Yeah that makes sense. A cross compiling section to the docs would be useful to document what you said.
And there's no harm in including it as a specialArg.

lib/default.nix
blaggacao: Well, there is [`pkgs-lib`](https://github.com/NixOS/nixpkgs/tree/master/pkgs/pkgs-lib) -- which is `lib` dependent on `pkgs`. And actually `build-support` would have been more accurately re-named as `pkgs-lib/build-support`, I guess, if it wasn't for historical reasons.

Concrete: I'd want to make a library function that is not pure nix, but uses any other scripting language.

I was thinking that I probably _could_ instatiate nixos and get access to `pkgs`  but that would be a different `pkgs` than the rest of this repository.

Pacman99: I see your point and I like the idea of library functions written in other languages. 
But I would argue that the lib folder in devos root, should just be pure nix functions that are used across devos just like nixpkgs's lib. 
And for other scripts or library functions using other languages belong as an overlay or in pkgs/. As they need to be derivations.
The reason I brought this up, is if lib gets an instance of pkgs, it would always get x86_linux and that would break full multi-arch.

Pacman99: And following what you said, perhaps you could add a pkgs-lib to the pkgs folder. That would would get multi-arch support too.

blaggacao: @Pacman99 You're absolutely spot on with your reply / analysis. Let's drop this then and I'll refactor the docs and adapt my use case.

blaggacao: https://github.com/divnix/devos/issues/166

hosts/NixOS-a64.nix
Pacman99: Yeah a test would be a good idea, I'll try to add one. 
But we still need to communicate to users that devos will respect nixpkgs.system without them having to dig through hosts/default.nix.

hosts/README.md
```diff
@@ -25,6 +25,10 @@ that you intend to use on your machine.
 Additionally, this is the perfect place to import anything you might need from
 the [nixos-hardware][nixos-hardware] repository.
 
+You can set the option `nixpkgs.system` to indicates the host system's 
+architecture. This will result in the host getting an instance of nixpkgs for 
+that architecture. The definition must be one exported by the nixpkgs flake.
```
Pacman99, line 30: Does this sound ok?

hosts/NixOS-a64.nix
Pacman99: Will `nix flake check` fail if the tests you wrote were exported to all architectures? I was thinking of moving the checks output to under `systemOutputs` so the tests also have multi-arch support. I'm also trying to reduce the use of the one instantiated x86_64 pkgs above nixosConfigurations.

Pacman99: well I guess my question is if `nix flake check` tries to go through the checks for all architectures.

Pacman99: yup just tested it, looks like it does

hosts/README.md
blaggacao, line 30:
```suggestion
Set `nixpkgs.system` to the architecture of this host. Keep in mind 
that mostly not all packages are available for all architectures.
```

Pacman99: Yeah I like that better.

flake.nix
```diff
@@ -34,18 +34,17 @@
...
-          pkgs = pkgs'.${system};
+          defaultSystem = "x86_64-linux";
         in
         {
           nixosConfigurations =
             import ./hosts (nixos.lib.recursiveUpdate inputs {
-              inherit pkgs system extern;
-              inherit (pkgs) lib;
+              inherit multiPkgs defaultSystem extern;
+              lib = nixos.lib // { dev = lib; };
```
Pacman99, line 47: Not sure this is the best way to do this.

blaggacao: Have you seen this: https://github.com/divnix/devos/blob/d7f15a664d20de910611a29167107aee8ff0dbab/lib/devos/mkPkgs.nix#L53-L58

?

flake.nix
```diff
@@ -55,27 +54,28 @@
...
+        {
           checks =
             let
               tests = import ./tests { inherit self pkgs; };
-              deployChecks = builtins.mapAttrs
-                (system: deployLib: deployLib.deployChecks self.deploy)
-                deploy.lib;
+              deployHosts = nixos.lib.filterAttrs
+                (n: _: self.nixosConfigurations.${n}.config.nixpkgs.system == system) self.deploy.nodes;
+              deployChecks = deploy.lib.${system}.deployChecks { nodes = deployHosts; };
```
Pacman99, line 75: This isn't currently working. For some reason `nix flake check` evaluates all systems starting with 'x86_64'. Including x86_64-darwin, which now fails because the related nixpkgs doesn't evaluate on x86_64-linux.

Any ideas on what to do for this @nrdxp?

flake.nix
Pacman99: Yeah thats what is currently being used. But I'm trying to avoid instantiating a nixpkgs for x86_64, just for pure nix lib functions.

blaggacao: Maybe you could adopt it here:
```nix
 lib = nixos.lib.extend (lfinal: lprev: { 
   dev = self.lib;
   utils = inputs.utils.lib; 
 }); 
```

Pacman99:
Also switched to _module.args here, I think thats more correct.

Pacman99:
dropped multiPkgs as a module arg in favor of #164. This is a rare use case anyways, and self.packages can be used instead.

flake.nix
Pacman99: I'm trying to debug this and I can't understand why nix flake check evaluates x86_64-darwin checks. I would assume that it would only build the checks for the current system.

Pacman99: Ohh I see it tries to evaluate all systems which normally shouldn't be an issue, but because of the patched-nixpkgs it ends up using IFD. This is a known issue nixos/nix#4265.

Pacman99:
Fixed merge conflicts in latest push

blaggacao:
> And following what you said, perhaps you could add a pkgs-lib to the pkgs folder. That would would get multi-arch support too.

While starting to work on #166, would it be possible to create a folder `./lib/pkgs-lib` that _does_ depend on packages, but still lives within the `./lib` folder.

The reason is that I think `./lib` token face value interpretation should still prime over `./lib` as a dependency root, which to the average user is a little more subtle concept.

Hence the average user wold certainly wonder why a library function (regardless of it's dependencies) does not reside in a top level folder that is called precisely `lib`rary.

It's just an UX thing, but I think `devos` cares.

Pacman99:
I'm still against passing pkgs to lib/default.nix. It is consistent to use pkgs/pkgs-lib with the way nixpkgs does it so I don't see how it would be surprising or bad UX.

If the lib/pkgs-lib folder is imported in a seperate line in flake.nix thats ok. The issue right now is `lib` is not a multi arch output, because it should just be pure nix functions. So any function written with the pkgs passed to lib will only run on one architecture..

blaggacao:
> If the lib/pkgs-lib folder is imported in a seperate line in flake.nix thats ok.

That would have been my idea. I don't feel the nixpkgs's own top level dichotomy is particularly lucky. `./pkgs` is such a hodgepodge of all sorts of things. That makes it actually harder for beginners to cut through. It would be nice if we could rather git `./pkgs` a razor sharp mental profile.

> The issue right now is lib is not a multi arch output, because it should just be pure nix functions. So any function written with the pkgs passed to lib will only run on one architecture..

I'm goging to anticipate this for #166

Pacman99:
That is a good point, but I still think proper multi arch support is more important. Is it alright if I still remove it in this PR, since theres no function atm that needs it. And we can figure out how to reintegrate it properly when we need it.

blaggacao:
Perfect!

nrdxp:
> That is a good point, but I still think proper multi arch support is more important. Is it alright if I still remove it in this PR, since theres no function atm that needs it. And we can figure out how to reintegrate it properly when we need it.

Whe it's ready feel ready.

bors delegate+

Pacman99:
This will fail right now because of multi arch checks. So its blocked by #192.

hosts/NixOS-a64.nix
Pacman99: Dropped nixos-a64.nix. I'm not sure how to write a test specifically for multi arch. I might try to figure it out later, but I don't think its completely necessary. And mulltiarch checks helps too.

Pacman99:
I updated this PR so its no longer blocked by #192. Until we have proper multi-arch tests, the tests are now only imported if the system is x86_64-linux.

nrdxp:
Been out all day. Super exhausted, but I'll have a proper review of this tomorrow. Good job.

hosts/README.md
```diff
@@ -25,6 +25,9 @@ that you intend to use on your machine.
 Additionally, this is the perfect place to import anything you might need from
 the [nixos-hardware][nixos-hardware] repository.
 
+Set `nixpkgs.system` to the architecture of this host, default is "x86_64-linux".
```
nrdxp, line 28: I would make this a note. As the continuation from the previous paragraph doesn't really flow well. Plus, we want it to stand out so users don't overlook this information.

so Just
> ##### _Note_:
> Set `nixpkgs.system` to the architecture of this host, default is "x86_64-linux". ...

Other than that. I think this is ready

nrdxp (changes requested):
This is all but ready. Just fix my small nitpick in the docs and we can go ahead and merge it.

hosts/README.md
Pacman99: Good idea, that looks better. Can you look at the update. I followed the format of notes from the docs folder.

nrdxp: Almost there. The format I've used throughout the docs so far includes blockquotes and a level 5 header.

```markdown
> ##### _Note:_
> Some cool note.
```
or for multiple notes:

```markdown
> ##### _Notes:_
> * A note
> * Another note
```

Pacman99: Ok got it, should be fixed now

Pacman99: Missed the blockquotes. I think its good now.

nrdxp (approved):
Good to go.

bors r+

Issue #163, opened 2021-03-14
wireguard "wire-up"


blaggacao:
## Would your feature fix an existing issue?
Setting up site-to-site networking is hard.

## Describe the solution you'd like
A wireguard wire-up (in the accustomed well-though-through `devos` way) would make things a lot easier. It might be even conceivable that all hosts within a devos repository are wired-up by default and are immediately (even remotely) reachable over the wg interface and a pre-configured IP.

## Describe alternatives you've considered
`zerotier`

## Additional context
- https://christine.website/blog/my-wireguard-setup-2021-02-06

- In addition to wiring up one's own devos environment, this should also be helpful if anybody needs to contract help from the (`devos`) community for one of their environments. (like me contracting with somebody to help me with some harder things I'm hardly able to complete in due time on my own :wink:).
  - Such scenario could relatively cleanly be covered with documentation and maybe a configuration option which gives a consultant access to the all hosts under their wg IP address. (w/o handling of address conflicts, that might be a whole other beast)

Pacman99:
This would be really cool! I think one change in devos overall that could help is to add a specialArg that gives access to every other hosts configuration. Then we could add a module for wireguard that will generate the relevant configurations for each host.

nrdxp:
I was already planning on stealing the hosts.toml idea from this blog post. I was figuring that we could use it to setup deploy-rs nodes automagically, adding in the wireguard bits seem pretty awesome as well.

Pacman99:
Whats the point in adding a hosts.toml? Wouldn't it be better to use the module system to get information about each host? And if you need more information, you could just add some options.

nrdxp:
Now I see where you are going @Pacman99, your right, it may be better to pull the information directly from the other configs, to ensure information never goes stale.

blaggacao:
I also think `hosts.toml` can be rather an afterthought. This also goes alltogether into the direction of host identity. In  private repository I bootstrap hostidentity with spire/spiffe and also put the host onto the LDAP record (for a few other things such as nss-pam over the network). My reduced yaml config file looks like this for reference (of ideas what _can_ comprise a host identity):

```yaml
entries:
- spiffe_id: spiffe://playground/agents # alias for all spire agents in this cluster
  parent_id: spiffe://playground/spire/server
  selectors:
  - type: k8s_psat
    value: cluster:playground
  - type: k8s_psat
    value: agent_ns:spire
  - type: k8s_psat
    value: agent_sa:spire-agent
[...]

PR #164, opened 2021-03-14
add hosts module arg

### LINKED ISSUE #169: Pass 'self' to lib
For os.mkPkgs and os.mkPackages, self is passed to them as a function argument. But why not just pass self to lib/default.nix and make it available for all lib functions? 
So you can just call `os.mkPkgs` without any arguments. And `os.mkPackages` can just be passed `pkgs`.

Pacman99:
should help with #163. Fixes #169
But either way this could be generally useful. I have one use case of setting up a minecraft bungeecord proxy with servers on different hosts.

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -75,6 +75,8 @@ let
 
         _module.args = {
           inherit self;
+          hosts = builtins.mapAttrs (_: host: host.config)
+            (removeAttrs hosts [ hostName ]);
         };
       };
     in


nrdxp:
This looks fairly benign, perhaps a usage example in the docs would be pertinent

hosts/default.nix
```diff
@@ -73,6 +71,12 @@ let
 
         networking = { inherit hostName; };
       };
+
+      specialArgs = extern.specialArgs // {
+        inherit suites;
+        hosts = builtins.mapAttrs (_: host: host.config)
+          (removeAttrs hosts [ hostName ]);
```
Pacman99, line 78: Not sure if the current host needs to be removed

Pacman99:
Agreed a usage example would be nice. I'm not sure where to put one. This change could be useful for profiles, modules, and hosts. But I guess hosts would still be the best place to include it.

nrdxp:
Actually, I was wanting to add a section to the docs, something like `Common Usecases` or something to that effect. Perhaps it would be best there.

hosts/default.nix
Pacman99: I took out the line, hosts is now every host's config including the current.

Pacman99:
Yeah that would be perfect

hosts/default.nix
nrdxp: I would say to probably leave it how you had it originally, as a system can already access it's own config from the `config` module arg.

Pacman99: Yeah thats fair.

Pacman99:
If we do add `multiPkgs` in the other PR, we could document both as special arguments available to hosts.

Pacman99:
switched to use _module.args based on comment in nixpkgs/lib/modules.nix. hosts arg should never be used in an imports line - evaluated modules can't be imported again,.

Pacman99:
I increased the scope of this PR to cover #169. Now lib functions can access self and any input. And inputs is passed as a special arg, since that could be used in an imports statement, while self is passed as a module arg.

hosts/default.nix
```diff
@@ -72,6 +73,12 @@ let
         ];
 
         networking = { inherit hostName; };
+
+        _module.args = {
+          inherit self;
```
Pacman99, line 78: I know `devos` was suggested for this. I don't see much harm in just using self, I don't think thats used anywhere else. And self is correct.

hosts/default.nix
```diff
@@ -72,6 +73,12 @@ let
...
+          hosts = builtins.mapAttrs (_: host: host.config)
+            (removeAttrs hosts [ hostName ]);
```
Pacman99, line 80: Leaving this still as a convenience. Its much easier to map over this instead of self.nixosConfigurations.*.config. And it removes the current config - useful for creating config files that only need to include other hosts.

nrdxp: LGTM

hosts/default.nix
blaggacao: I agree, with flakes, the semantic of `self` ia shifting. Similarily, overlays now enforce `prev` & `final` instead of former `self` & `super` convention.

blaggacao:
LGTM

I think it might be still too early for a smdocumented usage example. As I understand it, this idea came up towards a possible way reacting to another hosts configuration. That is mostly it's identity (hostname, ip, systemd machine id, &mdash; maybe static link local ipv6?)
r

Pacman99:
Yeah exactly, although with self and inputs I think this PR has more value. It is now an alternative method of passing inputs to your hosts instead of extern. So for mobile-nixos you can't use the extern modules section, because you can't add the pinephone module to all hosts. So using the inputs specialArg would be better. You can also use self.packages for cross compiling.

I don't know much about wireguard, but I would imagine that you might need to know information about other hosts to create the config for each host.

blaggacao:
I carved out the `self` part in #189 so that we can reason about `inputs` and `hosts` independently and also to unblock #168

Pacman99:
I think either way we're blocked due to the ci issues. And I think this PR is pretty much ready to merge.

flake.nix
```diff
@@ -57,7 +57,7 @@
           overlay = import ./pkgs;
           overlays = lib.pathsToImportedAttrs (lib.pathsIn ./overlays);
 
-          lib = import ./lib { inherit nixos pkgs; };
+          lib = import ./lib (inputs // { inherit nixos pkgs self; });
```
Pacman99, line 60: I was debating over doing this or just making it available as `inputs`. This way a lib function could just do something like `{ nix-darwin, ...}:` and get access to nix-darwin.

Pacman99: And also a lib function could do `args@{ ... }:` to get access to all inputs. This is probably only useful for #190

blaggacao: Maybe let's keep it name spaced as `inputs`. Similar to `self` this is well-known self-explaining top level flake glossary.

Pacman99: I personally prefer this method because it feels more in alignment with the `callPackage` paradigm - or in this case `callLib`. You specify the things you need for your function and you get those inputs. If we just pass inputs itself, then you take everything, which is more similar to the module system. 
But I'm ok with either, they both get the job done.

blaggacao: hm, for the sake of consistency, we probably shouldn't import it differently than how we import it with `./hosts`, though.

Pacman99:
So this PR now only covers passing inputs to lib and hosts. And it adds the hosts module arg as a convenience.

flake.nix
Pacman99: > hm, for the sake of consistency, we probably shouldn't import it differently than how we import it with ./hosts, though.

But do we really need consistency between lib and hosts? they are two different things.

@nrdxp Which one would you prefer? I think once decide how to pass inputs to lib, this should be ready.

blaggacao:
Do we actually want to pass `inputs` to `./hosts`? see https://github.com/divnix/devos/issues/190#issuecomment-800610201

Pacman99:
My reasoning is theres no harm in doing so. And it helps with flakes containing nixosModules that you only want in one host.

blaggacao:
> And it helps with flakes containing nixosModules that you only want in one host.

Is this an issue despite lazy evaluation? (sorry my ignorance)

> And extern doesn't cover all use cases with inputs 

Did you have any specific use case in mind?

> My reasoning is theres no harm in doing so.

Well, it dilutes the interface. I don't think that's good on such a general basis which affects `./profiles` and `./modules` downstream. Or at least we should have a clear picture of what our interfaces are here.

Pacman99:
I removed that part. I realize its a pretty niche use case and also I can just do it manually in `extern` with an `inherit inputs;`.

flake.nix
blaggacao: `nix.registry = mapAttrs (n: v: { flake = ${v}; }) inputs;` look clearer, indeed :wink: 

No need to be aware if some additionaly unwanted attributes sneak in and cause nasty side effects in unexpected places.

blaggacao:
Thank you @Pacman99 for all this inspiration and thought fodder.

Interaction / discussion has it all! :heart:

Pacman99:
> Thank you @Pacman99 for all this inspiration and thought fodder.
> Interaction / discussion has it all! heart

Indeed it does! Thank you too!

Pacman99:
Also this is back to its original state, just adds hosts module arg. So I think its ready to merge.

nrdxp:
bors delegate+

nrdxp (approved):
bors r+

PR #165, opened 2021-03-15
doc: identify installation device

blaggacao:
Is this the best tool to check the `/dev/sd*`s & `/dev/mblk*`s, etc.?

It is, in any case, available in the nixos installer iso.

### DIFF
--- a/doc/start/iso.md
+++ b/doc/start/iso.md
@@ -8,4 +8,7 @@ dd bs=4M if=result/iso/*.iso of=/dev/$your_installation_device \
   status=progress oflag=sync
 ```
 
+_Note: You can use `lsblk` to better identify `$your_installation_device` and
+it's current partitions._
+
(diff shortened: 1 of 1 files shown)


doc/start/iso.md
```diff
@@ -8,4 +8,7 @@ dd bs=4M if=result/iso/*.iso of=/dev/$your_installation_device \
   status=progress oflag=sync
 ```
 
+_Note: You can use `lsblk` to better identify `$your_installation_device` and
```
nrdxp, line 11: Note format everywhere else in the docs is:
```markdown
> ##### _Note:_
> My awesome note
```

I know I've been on certain minimal systems where lsblk hasn't been available, so perhaps we should include utillinux as a devshell dependency? Maybe that's overdoing it though.

nrdxp (changes requested):

Issue #152, opened 2021-03-05
Toward a stable filesystem api


nrdxp:
## Second Draft
We need to adapt to Nix now being a language with a proper module system, considering flakes to be essentially equivalent to a module in other languages. This should help create clear and distinct lines between different elements of the system, and keeps private code from leaking. For example, profiles could be it's own subflake of DevOS, which exports all profiles as `nixosModules` to then be consumed by the parent flake. Not every directory need be a flake, and many of the existing directory folders could be grouped together in a shared flake to reduce the complexity of the filesystem layout, without sacrificing any of the already established conveniences of DevOS. 

This has the added benefit of allowing power users to pull only relevant pieces of DevOS, such as it's lib, in a flake ref: `"github:divnix/devos?dir=lib"`.

## Impetus
In a word: complexity.

From the very first moment of deploying DevOS to the world, I knew the fs hierarchy had to change. I was hoping to illicit feedback from the community. As an example, one of the complaints I hear echoed is the default imports, which happen it different ways in different places, and if kept at all should definitely be reduced to a simpler implementation. This issue will remain open until a solid layout has been decided on, so we don't have to continue changing it out from under users. I want a flake centric approach, and internal flakes seem like the right way to manage access, at least from a high level. 

For example, with this approach we should be able to maintain the convenience of the current doc setup, i.e a README.md in every directory explaining it's use, without actually needing the book.toml and SUMMARY.md in the root. We can just pull in the `src` flake to the `doc` flake, and overlay the book.toml onto it to generate the book.

Once a demo is achieved that maintains all the established conveniences of current DevOS, and a proof-of-concept for #127, I will open a PR that we can then scrutinize.

## Changes
We would turn off the github template, and generate the template as a derivation with the docs optionally included for users. This will eliminate one level of nesting for users, while also eliminated irelevant code from the users template.

## Goal
This won't change once 1.0 is reached, so we need to get it right, aiming at convenience for users without restricting what they can do.

Pacman99:
Can you explain what you mean by out? And why things like overlays and pkgs go there?

And I'm not sure how to fix this or if it should be changed, but its pretty annoying that to edit a profile you now have to go three directories down(including profile folder). I feel like profiles are the most used folder. 

But overall I do like the separation of different folders.

nrdxp:
[...]

PR #168, opened 2021-03-15
[Merged by Bors] - lib/devos: bake devos repo into live cd

### LINKED ISSUE #167: doc: quickstart "ISO. What next?"
Given one started to plan an environment with devos:

There is a bit of a gap between flashing the ISO of a new host onto a stick and fully managing a host either locally (after having set up network and access rights and cloned the repo onto it) or remotely via deployrs (after having bootstrapped identity and ssh access).

Related: https://github.com/divnix/devos/issues/132#issuecomment-799016542

documenting my journey so far:

1. `flk iso MyNewHost`
2. `dd ...`
3. boot into iso on target host
4. follow https://nixos.org/manual/nixos/stable/index.html#sec-installation-partitioning
5.  1), 2) & 3) of https://nixos.org/manual/nixos/stable/index.html#sec-installation-installing, that is: 
  - `mount /dev/disk/by-label/nixos /mnt`
  - `mkdir -p /mnt/boot`
  - `mount /dev/disk/by-label/boot /mnt/boot`
  - `swapon /dev/sda2`
6. keep target running off of the usb memory card
7. from original machine: `deploy ".#MyNewHost" --hostname 192.168.191.179 --ssh-user nixos` & that failed, presumably because according to nixos installer things are to be mounted under `/mnt`

blaggacao:
fix #167 

This worked for me to bootstrap another machine.

### DIFF
--- a/doc/start/iso.md
+++ b/doc/start/iso.md
@@ -9,3 +9,36 @@ dd bs=4M if=result/iso/*.iso of=/dev/$your_installation_device \
 ```
 
 This works for any file matching `hosts/*.nix` excluding `default.nix`.
+
+## Bootstrap Target Machine
+
+Once your target host has booted into the live iso, you need to partion
+and format your disk according to the [official manual][manual].
+
+Then properly mount the formatted partitions at `/mnt`, so that you can
+install your system to those new partitions.
+
+Mount `nixos` partition to `/mnt` and &mdash; for UEFI &mdash; `boot` 
+partition to `/mnt/boot`:
+
+```console
+$ mount /dev/disk/by-label/nixos /mnt
+$ mkdir -p /mnt/boot && mount /dev/disk/by-label/boot /mnt/boot # UEFI only
+$ swapon /dev/$your_swap_partition
+```
+
+Install using the `flk` wrapper baked into the iso off of a copy of devos 
+from the time the iso was built:
+
+```console
+$ cd /iso/devos
+$ nix develop
+$ flk install NixOS --impure # use same host as above
+```
+
+_Note: You _could_ install another machine than the one your iso was built for,
+but the iso doesn't necesarily already carry all the necesary build artifacts._
+
+<!-- TODO: find out why --impure is necesary / PRs welcome! -->
+
+[manual]: https://nixos.org/manual/nixos/stable/index.html#sec-installation-partitioning
--- a/doc/start/iso.md
+++ b/doc/start/iso.md
@@ -10,10 +10,67 @@ dd bs=4M if=result/iso/*.iso of=/dev/$your_installation_device \
 
 This works for any file matching `hosts/*.nix` excluding `default.nix`.
 
+## Remote access to the live installer
+
+The iso live installer comes preconfigured with a network configuration
+which announces it's hostname via [MulticastDNS][mDNS] as `hostname.local`, 
+that is `NixOS.local` in the above example.
+
+In the rare case that [MulticastDNS][mDNS] is not availabe or turned off
+in your network, there is a static link-local IPv6 address configured to
+`fe80::47`(mnemonic from the letter's position in the english alphabet:
+`n=14 i=9 x=24; 47 = n+i+x`).
+
+Provided that you have added your public key to the authorized keys of the
+`nixos` user:
+
+```nix
+{ ... }:
+{
+  users.users.nixos.openssh.authorizedKeys.keyFiles = [
+    ../secrets/path/to/key.pub
+  ];
+}
+```
+
+You can then ssh into the live installer through one of the 
+following options:
+
+```console
+ssh nixos@NixOS.local
+
+ssh nixos@fe80::47%eno1  # where eno1 is your network interface on which you are linked to the target
+```
+
+_Note: the [static link-local IPv6 address][staticLLA] and [MulticastDNS][mDNS] is only
+configured on the live installer. If you wish to enable [MulticastDNS][mDNS]
+for your environment, you ought to configure that in a regular [profile](../../profiles)._
+
+## EUI-64 LLA & Host Identity
+
+The iso's IPv6 Link Local Address (LLA) is configured with a static 64-bit Extended
+Unique Identifiers (EUI-64) that is derived from the host interface's Message
+Authentication Code (MAC) address.
+
+After a little while (a few seconds), you can remotely disvover this unique and host
+specific address over [NDP][NDP] for example with:
+
+```console
+ip -6 neigh show # also shows fe80::47
+```
+
+***This LLA is stable for the host, unless you need to swap that particular network card.*** 
+Under this reservation, though, you may use this EUI-64 to wire up a specific
+(cryptographic) host identity.
+
 ## Bootstrap Target Machine
+
+_Note: nothing prevents you from remotely exceuting the boostrapping process._
 
 Once your target host has booted into the live iso, you need to partion
 and format your disk according to the [official manual][manual].
+
+### Mount partitions
 
 Then properly mount the formatted partitions at `/mnt`, so that you can
 install your system to those new partitions.
@@ -27,6 +84,8 @@ $ mkdir -p /mnt/boot && mount /dev/disk/by-label/boot /mnt/boot # UEFI only
 $ swapon /dev/$your_swap_partition
 ```
 
+### Install
+
 Install using the `flk` wrapper baked into the iso off of a copy of devos 
 from the time the iso was built:
 
@@ -36,9 +95,14 @@ $ nix develop
 $ flk install NixOS --impure # use same host as above
 ```
 
-_Note: You _could_ install another machine than the one your iso was built for,
-but the iso doesn't necesarily already carry all the necesary build artifacts._
-
 <!-- TODO: find out why --impure is necesary / PRs welcome! -->
 
+_Note: You _could_ install another machine than the one your iso was built for,
+but the iso doesn't carry all the necesary build artifacts so the target would
+start to build the missing parts on demand instead of substituting them from
+the iso itself._
+
 [manual]: https://nixos.org/manual/nixos/stable/index.html#sec-installation-partitioning
+[mDNS]: https://en.wikipedia.org/wiki/Multicast_DNS
+[NDP]: https://en.wikipedia.org/wiki/Neighbor_Discovery_Protocol
+[staticLLA]: https://tools.ietf.org/html/rfc7404
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -22,6 +22,34 @@ lib.nixosSystem (args // {
               networking.networkmanager.enable = lib.mkForce false;
               # confilcts with networking.wireless
               networking.wireless.iwd.enable = lib.mkForce false;
+              # Set up a link-local boostrap network
+              # See also: https://github.com/NixOS/nixpkgs/issues/75515#issuecomment-571661659
+              networking.usePredictableInterfaceNames = lib.mkForce true; # so prefix matching works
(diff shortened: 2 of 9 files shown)


lib/devos/devosSystem.nix
```diff
@@ -13,6 +13,10 @@ lib.nixosSystem (args // {
             "${nixos}/${modpath}/${cd}"
             ({ config, ... }: {
               isoImage.isoBaseName = "nixos-" + config.networking.hostName;
+              isoImage.contents = [{
+                source = ../..;
```
blaggacao, line 17: Is there a thought construct here that would command us to better use `inputs.self` ?

doc/start/iso.md
```diff
@@ -9,3 +9,36 @@ dd bs=4M if=result/iso/*.iso of=/dev/$your_installation_device \
...
+```console
+$ cd /iso/devos
+$ nix develop
+$ flk install NixOS --impure # use same host as above
+```
+
+_Note: You _could_ install another machine than the one your iso was built for,
+but the iso doesn't necesarily already carry all the necesary build artifacts._
+
+<!-- TODO: find out why --impure is necesary / PRs welcome! -->
```
blaggacao, line 42: I was stumbling on and over `access to path '/mnt/nix/...' is forbidden in restricted mode`

blaggacao (commented):
Two specific doubts, and a general one:

Are store paths actually copied over from the iso?

I couldn't tell from: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-install.sh#L143

blaggacao:
I think we do encounter a problem with secrets, though. It appears to me they are copied over as binaries and not quite useful, hence. This is only really useful for some basic bootstrapping.

Please review, but do not merge. I think we need a [better host bootstrapping](https://github.com/serokell/deploy-rs/issues/74).

lib/devos/devosSystem.nix
blaggacao: continue: https://github.com/divnix/devos/issues/169

blaggacao:
I think I was tired already yesterday. Resolved the previous comment. It was an unrelated issue.

nrdxp:
This is looking pretty good. Have your considered all adding [`isoImage.includeSystemBuildDependencies`](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/cd-dvd/iso-image.nix#L479-L488). It may help with offline installation.

blaggacao:
> This is looking pretty good. Have your considered also adding `isoImage.includeSystemBuildDependencies`. It may help with offline installation at the cost of a bigger image.

Cool! I'll do a couple of tests today, still, and will include that.

I also want to experiment to _also_ set a static link local ip on the network interface when using the live cd. That way, we can feed a well known ip to the `deploy` when the live cd is running: so a technician can plug in the usb stick and a devops can do the depoyment "remotely" (if they are on the same link local network). At least the devops doesn't have to crawl around under the office desks.

Pacman99:
Isn't hostname usually enough for that? I'm not sure about adding a static ip, because of possible ip conflicts and other issues that can happen. Worst case you could find the ip of the machine by searching by hostname.

blaggacao:
> Isn't hostname usually enough for that?

Host name needs identity (which we actually have here since we build bespoke iso installers, but we also might want to work with a generic hostconfig, instead, eg. `NixOS`) and also dns to attest / route identities.

I still want to cover the case of an anonymous bootstrap image / iso that we can burn onto stockpiled computers and then ship to the site and initiate with an identity over the network.

Pacman99:
I definitely understand the benefits of static IP. I've just had issues with it on my network and I know it can be kind of finicky. And I don't want the machine to lose network connection entirely for the convenience of a static ip.

blaggacao:
> I've just had issues with it on my network and I know it can be kind of finicky. And I don't want the machine to lose network connection entirely for the convenience of a static ip.

I think that is the exact use case for link local ad hoc networks
>    IPv4 Link-Local addresses are not suitable for communication with
   devices not directly connected to the same physical (or logical)
   link, and are only used where stable, routable addresses are not
   available (such as on ad hoc or isolated networks)

https://tools.ietf.org/html/rfc3927

So we should be good.

Pacman99:
Also I was thinking that we should extract the custom system builds outside of lib, so users can add their own custom builds for all their hosts. The idea behind devosSystem is really great and I found a use for it in the home manager PR, and I'm sure others might have good ideas. And they can customize the existing systems we configure.

Pacman99:
> I think that is the exact use case for link local ad hoc networks

I'll have to learn more about that. But it looks like you have a good plan!

blaggacao:
> `isoImage.includeSystemBuildDependencies`

It looks as if this easily exceeds the size of an usb flash drive.

But foremost, I hope that [`auto?trusted=1` substituter](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-install.sh#L143) actually let's the bespoke iso image act as a substituter. Since the repo baked into it is the same as the moment the image was created, I suppose that all (or almost all) builds are copied over from the installer iso and no build is done on the target.

nrdxp:
It looks like this is getting pretty close. I'm going to delegate this to both of you, while I work on opening a further PR toward #152. There is no harm in trying a `bors r+` to merge, but it will fail til I fix the CI runner. I have a few more emails to respond to for the day, and then I can get it fixed up. Good work.
bors delegate+
bors delegate=@Pacman99

blaggacao:
I'm checking out [MulticastDNS](https://tools.ietf.org/html/rfc6762#page-5) as an alternative to static IPs. that might be slightly friendlier to the end user, althouth I'm not sure if that depends on fancy router support...

that would look something like `deploy ... bare-host.local.` or `deploy ... bootstrapping.local.`

blaggacao:
Work in progress: (feedback welcome)

```nix
{
              systemd.network = {
                enable = true;
                # See also: https://github.com/NixOS/nixpkgs/issues/75515#issuecomment-571661659
                # https://www.freedesktop.org/software/systemd/man/systemd.network.html
                networks."boostrap-link-local" = {
                  matchConfig = {
                    Name = "en* wl* ww*";
                  };
                  networkConfig = {
                    Description = "Link-local host bootstrap network";
                    MulticastDNS = true;
                    LinkLocalAddressing = "ipv6";
                  };
                  extraConfig = ''
                    # Unique, yet stable. Based off the MAC address.
                    IPv6LinkLocalAddressGenerationMode = "eui64"
                  '';
                };
              };
}
```

doc/start/iso.md
```diff
@@ -9,3 +9,103 @@ dd bs=4M if=result/iso/*.iso of=/dev/$your_installation_device \
...
+From within the devshell:
+
+```console
+deploy "flk#NixOS" --ssh-user nixos --hostname NixOS.local # or, if multicastDNS fails:
+deploy "flk#NixOS" --ssh-user nixos --hostname [fe80::47%eno1]
+```
+
+The latter option is currently blocked by [Nix Issue 4643](https://github.com/NixOS/nix/issues/4643).
+
+TODO: The former did not complete on my machine as the MulticastDNS seemed to be too unstable.
```
blaggacao, line 105: blocked by https://github.com/NixOS/nix/issues/4643

The `NixOS.local` MulticastDNS didn't seem to be utterly stable on my network and disappeared rather quickly (between the deploy's nix copy &mdash; suceeded &mdash; and remote activation &mdash; failed)

lib/devos/devosSystem.nix
Pacman99: You could rebase this off #164 so you can directly get and pass self

lib/devos/devosSystem.nix
```diff
@@ -13,11 +13,43 @@ lib.nixosSystem (args // {
...
+                target = "/devos/";
+              }];
               # confilcts with networking.wireless which might be slightly
               # more useful on a stick
               networking.networkmanager.enable = lib.mkForce false;
               # confilcts with networking.wireless
               networking.wireless.iwd.enable = lib.mkForce false;
+              # Set up a link-local boostrap network
```
Pacman99, line 25: What do you think about separating the fancy networking parts into another build? Since some of that is experimental and also has a different goal. So like isoConfig could just be a simple iso with the necessary dependencies to install a new system through a usb. And another config for the path to bootstrapping machines over the nework.

blaggacao: I made sure that the default behavior is covered via `DHCP = true` for all network interfaces that start with `wl` , `ww` & `en` (with predictable names force enabled).

That seems to be all physical network devices. I assume base line conmectivity, thus, is resonably secured across devices.

Assuming that we have our backs covered, the goal of consistent amd predictable addressing of the live installer is first and foremost to be able to execute all installation routines (like partitioning & formatting) remotely.

That means, among other things, a user does not even have to connect a keyboard and a screen to the device they want to boostrap.

I don't find this use case sufficiently distinguishable to afford a different usb stick (one eith remote access and one without).

blaggacao:
I'll do another manual test, cleanup & rebase today, so we should be close to ".good to go".

lib/devos/devosSystem.nix
Pacman99: Sounds good! Yeah I thought about it a bit more and I agree with you.

Pacman99 (approved):

Pacman99:
This looks really good. I think the only thing I would like to change is to use `self` instead of `../..` to reference the flake. But that could be done separately too.

Pacman99:
Also I had another idea for the live cd. You could include all the inputs in the ISO's registry. So while installing you don't need to download the inputs again. Once all inputs get passed to lib you could do something like:
```
nix.registry = mapAttrs (n: v: ${n}.flake = ${v}) inputs;
```
This would also require adding `inputs@{...}` to the top.

But not sure this would work, and probably best to save it for a future PR.

blaggacao:
Lat's call this a day! We might consider opening issues / discussions for further improvements:

- use ipv4 link-local to tie it to a cachix derviation that represents the host's profile and let the stick autoprovision itself (unattended mode).
- put the inputs in the registry so that they don't need downloading (aka. aire gapped installation &mdash; to be checked if that is actually enough)

Pacman99 (approved):

Pacman99:
Ohh right this is going to do a squash merge :(

blaggacao:
> Ohh right this is going to do a squash merge :(

Never mind, I think Tim is going to change that soon...

nrdxp:
sorry for delay, I believe I figured out how to fix the ci runner while keeping everything up to date. Essentially we will use the stable nix daemon, and only rely on nixosUnstable for the client side. On my tests last night this seemed to fix things, but a lot of derivations were failing. Bors won't commit with failing tests, so I'll see what I can do to fix them, or we may just have to disable some of the failing archtiectures for a while.`

blaggacao:
@nrdxp If you exceptionally can merge this manually, we would save my carefully crafted git history for once. :wink: 

I built everything on my machine and tested end to end on target host.

Only caveat: `deploy` doesn't know how to sudo, but this needs improvment in `doc/integrations/deploy.md` in a separate PR.

nrdxp:
Well, give me about half an hour to implement the fixes in the CI runner, and if it still doesn't work, I'll just merge manually.

nrdxp:
So it merged. There is the `bors r-` to cancel a merge before tests finish for future reference. I just removed squash merge manually, but it was too late.

Pacman99:
Yeah apologies, I will remember that for the future.

nrdxp:
Not a big deal, it is inevitable with any workflow shift to have a few kinks to work out. I am glad that @blaggacao at least got credit in the commit message though.

blaggacao:
I'm not after any credit whatsoever :wink:

PR #189, opened 2021-03-16
general: pass `self` to `./lib` & `./hosts`

### LINKED ISSUE #169: Pass 'self' to lib

blaggacao:
It is generally useful to acess the top level flake from
library functions or hosts. This not only simplifies
the mental model and code but also provides additional
context and not least a handle to the repo source code
in the nix store.

closes #169

bors delegate=@Pacman99

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -1,14 +1,14 @@
-args@{ nixos, pkgs, ... }:
+args@{ nixos, pkgs, self, ... }:
 let inherit (nixos) lib; in
-lib.makeExtensible (self:
+lib.makeExtensible (final:
   let callLibs = file: import file
     ({
       inherit lib;
 
-      dev = self;
+      dev = final;
     } // args);
   in
-  with self;
+  with final;
   {
     inherit callLibs;
 
--- a/flake.nix
+++ b/flake.nix
@@ -36,7 +36,7 @@
 
       extern = import ./extern { inherit inputs; };
 
-      pkgs' = os.mkPkgs { inherit self; };
+      pkgs' = os.mkPkgs;
 
       outputs =
         let
@@ -57,7 +57,7 @@
           overlay = import ./pkgs;
           overlays = lib.pathsToImportedAttrs (lib.pathsIn ./overlays);
 
-          lib = import ./lib { inherit nixos pkgs; };
+          lib = import ./lib { inherit nixos pkgs self; };
 
           templates.flk.path = ./.;
           templates.flk.description = "flk template";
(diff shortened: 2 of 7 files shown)


blaggacao:
bors try

blaggacao:
@Pacman99 Is this clarifying force push correct?

https://github.com/divnix/devos/compare/0173d29843c5f272a11bbec9ea56f938f759471c..30f18c720e917f71caf12e78084b0ab7c3a09707

lib/default.nix
```diff
@@ -1,14 +1,14 @@
-args@{ nixos, pkgs, ... }:
+args@{ nixos, pkgs, self, ... }:
 let inherit (nixos) lib; in
-lib.makeExtensible (self:
+lib.makeExtensible (selfLib:
```
Pacman99, line 3: This can actually stay as self since the 'self' is a part of args and is irrelevant for the lib/default.nix. And function arguments take precedence.

Pacman99:
Yeah this makes sense, it would be good to get in these changes soon.

lib/default.nix
Pacman99: and I don't think it really hurts readability

blaggacao: Yea, but it's very subtle and just really confusing and not very nice to read.

blaggacao: We'll I was like "wait, what's going on here!?!" &mdash; and we probably should reserve `self` for referring to flake top level. we could use `final` !?

Pacman99: Yeah I like final. Not a huge fan of `selfLib`.

Pacman99:
I think @nrdxp has to delegate one of us first since their the repo owner.

Pacman99 (approved):

nrdxp:
Is this a duplicate of #164?

blaggacao:
It is a carve out as direct dependency to #168 and is included in #168 .

I wanted to separate concerns of #164 in order to accentuate and walk through their use cases a little slower.

Pacman99:
I would prefer we discuss and merge #164 instead since thats already come up as a requirement for a couple other PR's. I don't see the point in picking apart the different parts of it into seperate PR's that need it.

nrdxp:
> I think @nrdxp has to delegate one of us first since they're the repo owner.

That is correct, delegation happens on a per PR basis.

blaggacao:
> I don't see the point in picking apart the different parts of it into seperate PR's that need it.

It's kind of a review tactic of mine... :wink: Also, #164 had a lot of force pushes which made it hard in my opinion to follow the complete thought process.

This was a dependency of #168 so it was just incorporated.

PR #191, opened 2021-03-16
iso: copy input closourse into iso to avoide re-download

### LINKED ISSUE #190: pin inputs into iso live registry
## Would your feature fix an existing issue?
With #168 , devos live iso gains the ability to enter the devos devshell and install through it on the target machine during bootstrapping.

Still, the target machine is fetching all inputs again.

By pinning inputs into the registry:
- the target will probably use them (if lockfile is not present and declares a different sha)
- more importantly, the nix store closures will get copied over (I hope I'm right?) and therefore there is no deed to fetch those files

n/a

## Additional context
- Suggested by @Pacman99 https://github.com/divnix/devos/pull/168#issuecomment-800420441
- similarily, we might want to find a way to copy the devshell closure to the iso, so that entering the devshell uses the baked in nix store closure.

blaggacao:
closes #190

Tested on my setup, considerably cuts times and only leaves one (noticable) download.

### DIFF
--- a/doc/start/iso.md
+++ b/doc/start/iso.md
@@ -46,6 +46,18 @@ _Note: the [static link-local IPv6 address][staticLLA] and [MulticastDNS][mDNS] 
 configured on the live installer. If you wish to enable [MulticastDNS][mDNS]
 for your environment, you ought to configure that in a regular [profile](../../profiles)._
 
+## ISO image nix store & cache
+
+The iso image holds the store to the live environment and _also_ acts as a binay cache
+to the installer. To considerably speed up things, the image already includes all flake
+`inputs` as well as the `devshell` closures.
+
+While you _could_ provision any machine with a single stick, a bespoke iso maximises
+those local cache hits.
+
+For hosts that don't differ too much, a common usb stick might be ok, whereas when
+there are bigger differences, a bespoke usb stick will be considerably faster.
+
 ## EUI-64 LLA & Host Identity
 
 The iso's IPv6 Link Local Address (LLA) is configured with a static 64-bit Extended
@@ -96,11 +108,6 @@ $ flk install NixOS --impure # use same host as above
 ```
 
 <!-- TODO: find out why --impure is necesary / PRs welcome! -->
-
-_Note: You _could_ install another machine than the one your iso was built for,
-but the iso doesn't carry all the necesary build artifacts so the target would
-start to build the missing parts on demand instead of substituting them from
-the iso itself._
 
 [manual]: https://nixos.org/manual/nixos/stable/index.html#sec-installation-partitioning
 [mDNS]: https://en.wikipedia.org/wiki/Multicast_DNS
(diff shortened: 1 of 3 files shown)


blaggacao:
@Pacman99 I tested this locally and it already works more than great. Thank you for the idea! Do you know how we could copy the devshell closure, too? Currently, the devshell seems to be the only remaining thing that is unecesarily built on the target machine.

blaggacao:
bors try

Pacman99:
ohh wait I actually haven't tested that code locally, I just wrote the code out. But it should be simple enough to build the isoConfig and inspect the result.

blaggacao:
> ohh wait I actually haven't tested that code locally, I just wrote the code out. But it should be simple enough to build the isoConfig and inspect the result.

I fully tried end to end and it works.

`nix registry list` shows all those local store paths.

Maybe we need to filter out the ones that are `flake = false, though`? or handle them differently? What do you think?

Pacman99:
Yeah you could, but it really wouldn't make much of a difference. Worst case theres just one unnecessary folder in the nix store.

blaggacao:
> Yeah you could, but it really wouldn't make much of a difference. Worst case theres just one unnecessary folder in the nix store.

Ok, let's don't bother then. That path might even be useful on the iso under some odd circumstances.

Pacman99:
So I'm not sure about this. But perhaps we could not import all modules of the system and just selectively pass along parts of the configuration. This would prevent starting up unnecessary system services while still adding necessary dependencies.

I think this would involve evaluating the modules in a seperate `nixosSystem` call. And just passing along some things. 

for example:
```nix
environment = { inherit (system.config.environment) systemPackages; };
```

blaggacao:
> So I'm not sure about this. But perhaps we could not import all modules of the system and just selectively pass along parts of the configuration. This would prevent starting up unnecessary system services while still adding necessary dependencies.

Do you refer to your previous suggestion to not enable all _modules_?

This PR is meant (and restrained to) specifically to target #190. I'll open an issue for that.

Pacman99:
Yeah its just another idea. Doesn't have to be a part of this PR.

blaggacao:
x-ref: https://github.com/divnix/devos/issues/194

nrdxp:
if it's ready, go ahead and merge
bors delegate+

nrdxp (approved):
LGTM

blaggacao:
I suspect that it actually does _not_ cache evaluations, since the commands still take some time in silence, that is no command output (which I only can explain with evaluation)?

Can those evaluations, like [this](https://hercules-ci.com/accounts/github/divnix/derivations/%2Fnix%2Fstore%2Fpncxk7zpm8vp4242g9413hsy6869y5i3-devshell.drv?via-job=ed571c00-c5a5-4ca8-a8af-719f2843302f), for example, also be copied into the store?

Pacman99 (approved):

PR #192, opened 2021-03-16
flake: nixos -> nixos-unstable

nrdxp:
20.09 is getting stale as we move toward a new release so track unstable for now.


profiles/core/default.nix
```diff
@@ -2,13 +2,14 @@
 let inherit (lib) fileContents;
 in
 {
-  nix.package = pkgs.nixFlakes;
+  nix.package = pkgs.nixStable;
```
Pacman99, line 5: Isn't it still `nixUnstable/nixFlakes`?

Pacman99:
Is it possible to drop the patched-nixpkgs now? That would allow for proper multi-arch checks.

profiles/core/default.nix
nrdxp: So this is the workaround for ci. Turns out, the new nix commands like flake and the rest are still able to function with a stable daemon as long as the command line client is unstable.

Pacman99: Ohh interesting! So how do you ensure the command line client is unstable?

profiles/core/default.nix
```diff
@@ -2,13 +2,14 @@
...
 
   nix.systemFeatures = [ "nixos-test" "benchmark" "big-parallel" "kvm" ];
 
   environment = {
 
     systemPackages = with pkgs; [
+      nixFlakes
```
Pacman99, line 12: Won't this mean both nix stable and unstable cli are installed?

profiles/core/default.nix
nrdxp: I added it to profiles/core `systemPackages`

profiles/core/default.nix
nrdxp: I thought so at first, but after trying it, it only installs unstable client thankfully.

Pacman99: Huh I wonder how the nix cli is installed by default in the first place... I would assume its set somewhere in nixpkgs. But as long as it work!

nrdxp:
failed derivations have to with:

```consile
test/tests/check/store/dg0fqn0xc0fxk4cr46mlbqry0gzvy2nw-dummy.drv':
             specified: sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
                got:    sha256-0qhPS4tlCTfsj3PNi+LHSt1akRumTfJ0WO2CKdqASiY=
make: *** [mk/lib.mk:106: tests/check.sh.test] Error 102
make: *** Waiting for unfinished jobs....
ran test tests/fetchGit.sh... [PASS]
```

which causes pretty much everything else to fail, since this is the only failing test, perhaps disabling tests on nixFlakes would be easiest solution. I'm gonna see if a subsequent patch will resolve it, but I suspect the patch will cause the hash to diverge again.

blaggacao:
I've seen that a few days ago. :smile:

profiles/core/default.nix
nrdxp: This will break nixos-unstable, since it sends the commands to the daemon,which doesn't know about flakes, but we workaround by using an alias for `nix shell <config> --comand switch-to-configuration <command> #switch boot etc`

nrdxp:
bors try

nrdxp:
I tried adding `doCheck = false` , but the tests are still being run and still causing a failure. A bit confused as to how to address this. Any ideas anyone?

Pacman99:
<details>
<summary>I'm getting this error while running `nix flake check` on this branch</summary>

```
error: builder for '/nix/store/63hfh6if3s0bp66vp2j6zimblv8ylhim-nix-2.4pre20210316_66fa1c7.drv' failed with exit code 2;
       last 10 log lines:
       >     rootCA (post)> + xargs nix copy --to file:///build/nix-test/tests/ca/substitute/binary_cache --no-require-sigs
       >     rootCA (post)> warning: you don't have Internet access; disabling some network-dependent features
       >     rootCA (post)> error (ignored): error: cannot unlink '/build/nix-test/tests/ca/substitute/binary_cache/nar/115r2bh5nm997p0g2a11ccpafx3rvb8ghvpf96zw6pl7zgzyla1x.nar.xz.tmp.20739': No such file or directory
       >     rootCA (post)> error: renaming '/build/nix-test/tests/ca/substitute/binary_cache/nar/115r2bh5nm997p0g2a11ccpafx3rvb8ghvpf96zw6pl7zgzyla1x.nar.xz.tmp.20739' to '/build/nix-test/tests/ca/substitute/binary_cache/nar/115r2bh5nm997p0g2a11ccpafx3rvb8ghvpf96zw6pl7zgzyla1x.nar.xz': No such file or directory
       >     error: program '../push-to-store.sh' failed with exit code 123
       > make: *** [mk/lib.mk:106: tests/ca/substitute.sh.test] Error 1
       > make: *** Waiting for unfinished jobs....
       > ran test tests/ca/build.sh... [PASS]
       > ran test tests/fetchMercurial.sh... [PASS]
       > ran test tests/flakes.sh... [PASS]
       For full logs, run 'nix log /nix/store/63hfh6if3s0bp66vp2j6zimblv8ylhim-nix-2.4pre20210316_66fa1c7.drv'.
error: 1 dependencies of derivation '/nix/store/z81ikz6qrazigmlrg2x8gj0wk1anw45s-devos-lib-tests.drv' failed to build
error: build of '/nix/store/11ga69pc8bh3km0ghiw9pryyaffa5xkd-jsonschema-deploy-system.drv', '/nix/store/6kdbkw2mf3a58l7f32wl3kg84jvvpbdx-deploy-rs-check-activate.drv', '/nix/store/dsv3dj0hc9qnh51b4wmm3q4sn7l2zaq6-vm-test-run-profiles.drv', '/nix/store/z81ikz6qrazigmlrg2x8gj0wk1anw45s-devos-lib-tests.drv' failed
```
</details>

Pacman99:
I wonder if its because of this:
`sed -i 's/experimental-features .*/& ca-derivations ca-references/' /build/nix-test/tests/ca/substitute/etc/nix.conf`
I'm not sure what it means, but if the test was meant to be done with specific nix.conf lines that are overriden by your patch it might fail.

Pacman99:
Also I don't understand the point of your custom patch? Whats wrong with just adding the experimental features in nix.conf?

nrdxp:
> Also I don't understand the point of your custom patch? Whats wrong with just adding the experimental features in nix.conf?

I understand how you feel. The main motivation was to avoid junk in the ci logs. I was wondering if there was a way to override global nix.conf with a user level nix.conf to remove the experimental-features line, accomplishing the same thing, but I don't think that ability exists, from what I could gather.

I may just have to remove it for now and deal with the junk.

Pacman99:
Got it, I assumed it had something to do with the ci, that makes sense. Thanks for clarifying. It is rather annoying to rebuild nix, but I understand if it helps with maintenance in any way.

nrdxp:
I had originally assumed their would be a way to set log level in nix.conf. Unless my grepping skills are lacking, I couldn't find one.

Pacman99:
I think dropping the custom patch might be the best way to move this forward. regular nixUnstable builds ok.

nrdxp:
I am attempting to get that together right now, but it reintroduced an old issue. I'd rather not resort to a shell script wrapper for nix, but it looks like that'll have to suffice, unless numtide/devshell#106

nrdxp:
interestingly, removing the patch didn't even resolve the build failures, so something else must still be going on.

Pacman99:
How come nix2.4 is still being rebuilt? Shouldn't it be pulled from the cache now if there are no custom patches?

hosts/default.nix
```diff
@@ -28,28 +28,44 @@ let
...
-        system.configurationRevision = lib.mkIf (self ? rev) self.rev;
-      };
+          nix.registry = {
+            devos.flake = self;
+            nixos.flake = nixos;
+            override.flake = override;
+          };
+
+          nix.extraOptions = ''
+            --experimental-features = ${lib.concatStringsSep " "
```
Pacman99, line 62:
```suggestion
            experimental-features = ${lib.concatStringsSep " "
```
is the `--` supposed to be there?

Pacman99:
If I pull nixUnstable from the nixos flake instead of nix, `nix flake check` passes for me. And it also prevents a rebuild of nix.

hosts/default.nix
nrdxp: I tink it's optional but it's nicer without

nrdxp:
Well I took your advice but ci is still failing as before.

Pacman99:
I can't reproduce locally this time either. nix flake check now passes for me on this branch..

Pacman99:
What happens if you update everything(including daemon) to nixUnstable like before? perhaps the previous issue you had with the ci was also because of the custom patch.

nrdxp:
bors try

nrdxp:
bors try

nrdxp:
> What happens if you update everything(including daemon) to nixUnstable like before? perhaps the previous issue you had with the ci was also because of the custom patch.

[this](https://hercules-ci.com/github/divnix/devos/jobs/157)
Surprisingingly, we finally have a few green builds, but almost everything is still failing.

Pacman99:
Even for job 157, this is whats confusing:
`FatalError {fatalErrorMessage = "Exception: opening a connection to remote store 'daemon' previously failed; type: nix::Error"} (worker: ExitFailure 1)` 
My guess would be that happens because of a daemon and cli mismatch.

nrdxp:
The thing that confuses me is that the previous lock file evaluated fine with hercules and the stable daemon unstable cli mistmatch. I am gonna try a few more things:
* see if perhaps it's a bad nixpkgs revision by updating the lock file once more
* putting hercules ci into a container where the both the cli and daemon on stable nix

blaggacao:
> see if perhaps it's a bad nixpkgs revision by updating the lock file once more

This was how I solved it locally. But I lost rack of any potential cause-effect relationships so didn't bother to recommend.

Pacman99:
>putting hercules ci into a container where the both the cli and daemon are stable nix - @Pacman99 this seems to be the solution

YAY!! I'm glad it worked. So this means the cli will be unstable while the nix daemon will be stable right? I thinks something we can work with.

profiles/core/default.nix
nrdxp: Okay, so maybe I was too tired yesterday and wasn't paying close attention, but I appear to have been wrong on this. I can't seem to get `nixFlakes` installed with `nix.package = nixStable` today. Not sure what the heck happened yesterday. Perhaps some of the [issues](https://discourse.nixos.org/t/aweful-nix-store-performance-on-zfs/11991) I've been having have something to do with it. I was able to get the desired behavior by adding nixFlakes to users.users.<myuser>.packages, but this isn't really a solution, as there doesn't seem to be a default way to add packages to each user.

nrdxp:
> > putting hercules ci into a container where the both the cli and daemon are stable nix - @Pacman99 this seems to be the solution
> 
> YAY!! I'm glad it worked. So this means the cli will be unstable while the nix daemon will be stable right? I think thats something we can work with.

Only problem is that the there is no easy to actually accomplish this without modifying the daemon version of the entire system. Perhaps I should modify the default nixos-container module on nixpkgs and add this functionality as a module option.

profiles/core/default.nix
Pacman99: Idk how much this helps, but when the package that gets passed to `nix.package` is then passed to systemd.packages. So thats how nix gets installed by default.

blaggacao:
I would actually be in favor of option 3 (hydra). We might just expect better interoperability guarantees during transition.

And there is also the flake "native" top level `jobs` attribute, which I don't fully underatand, but it smells flake-streamlined like rose water.

Pacman99:
I do really like hercules and you've put a lot of great work into setting it up. So I'm not sure either about switching to hydra. But I also don't know a lot about how either work, so I don't think I can really give a good argument for either side. Also could github actions be used with this: https://github.com/numtide/nix-flakes-installer?

I would like to avoid patching nixpkgs as that slows down evaluation a lot. And I'm not sure either change will be accepted upstream, as its more of a workaround than an actual solution. 

> Third option could be good, but I'm not sure. I think we would have to set up a website manually to view the CI results.

I thought this was included in hydra?

Pacman99:
Since the CI is still a WIP, could we just rely on `nix flake check` to determine if this is ready? I would say that is enough to ensure that devos will work for everyone. And we can work on setting up proper CI(hydra?) separately.
But at the same time that would break bors, so I'm not sure if we should. I believe the nix flakes installer I referenced above could be used as a temporary CI - it just runs nix flake check too. I'm not sure if github actions works with bors though.

nrdxp:
That sounds great to me. I'll just hold the CI machine back so it should still function as normal. I'll just have to try and resolve the merge conflicts. The only major issue was that CI is actually my main dev machine, as it's the beefiest machine I have access to ATM.

I just was trying to make sure I could keep up as a user. But I've pretty much worked around it, by just patching nixStable as `nix.package` on my machine, and installing nixUnstable as a part of my users profile. Since this is my PR, I'll go ahead and resolve the merge conflict and merge. Of course, if merging this breaks other PR's which are ready to merge. We can hold off until they are fully merged.

profiles/core/default.nix
Pacman99: if `nix.package` is set to nixFlakes, I don't think this is necessary anymore.

Pacman99:
> I just was trying to make sure I could keep up as a user. But I've pretty much worked around it, by just patching nixStable as nix.package on my machine, and installing nixUnstable as a part of my users profile. Since this is my PR, I'll go ahead and resolve the merge conflict and merge. Of course, if merging this breaks other PR's which are ready to merge. We can hold off until they are fully merged.

I would prefer to get this merged first. If anything this is the blocker for one of my PR's(home-manager). And I doubt dealing with merge conflicts from this would be that bad. 

I was looking at hydra, and it doesn't seem that hard to setup at all. There is a nixos module too and I believe the web interface is a part of hydra and it has github integration. We could open an issue here for sorting out CI and consider that a separate concern from upgrading devos. And until then rely on local runs of `nix flake check` for testing - which is realistically a very good way of testing.

Pacman99:
Also make sure to update the lock file as part of this PR, otherwise the `pkgs` flake hash is outdated and that breaks the flake.

profiles/core/default.nix
nrdxp: We are sticking with `nix.package = nixFlakes`.

hosts/default.nix
nrdxp: fixed

nrdxp:
I cleaned up the commit history quite a bit, since much of it was irrelevant and me just trying to fix CI. My only last question would be if we should update nixos-unstable since I started this PR about a week ago, but my vote would be no, since I've already built all the community package and uploaded them to cache.

I believe it is ready for final review @Pacman99.

Since I am manually patching my dev machine. CI and BORS should continue to work as normal even after this is merged.

Pacman99:
> My only last question would be if we should update nixos-unstable since I started this PR about a week ago, but my vote would be no, since I've already built all the community package and uploaded them to cache.

same here, a week difference in nixpkgs is really not much. And end users can update if they wish.

tests/default.nix
```diff
@@ -3,40 +3,10 @@ let
   inherit (self.inputs) nixos;
   inherit (self.nixosConfigurations.NixOS.config.lib) testModule specialArgs;
 
-  # current release 20.09 does not support the specialArgs required for us
-  # to use tests as we would normally use hosts. Using the "testing-python.nix"
-  # from the override flake would build the test-vm from an unstable os
-  # different than the one our systems are running. Instead simply patch nixpkgs
-  # to include the updated version. This can be removed in the next release
-  patchedNixpkgs =
```
Pacman99, line 11: I thought removing this would fix multi-arch checks, but it looks like it still fails to evaluate x86-darwin. I think its because of passing system to the make-test-python function. Not really relevant to this PR, but tests could use a bit of reworking to improve multi-arch support.

Pacman99 (approved):
This passes nix flake check on my system and the changes look great.

tests/default.nix
nrdxp: This may not be a huge deal as darwin is currently disabled in ci for similar reasons. I'd like to invest in a decent mac at some point and run a separate CI runner on it to build these packages. For now we should probably filter out of checks as we do with CI.

Issue #195, opened 2021-03-16
copy evaluation store paths to iso


blaggacao:
If you, by-comer, are knowledgeable enough to solve https://github.com/divnix/devos/pull/191#issuecomment-800685936, please do.

The goal is to also copy evaluations into the iso to avoid evaluation times on slow/low resource target machines.

Pacman99:
I'm not sure if I misunderstand this. But no matter how many paths we add to the nix store by default, nix will still have to evaluate the nix expressions to figure out what store path it needs. And so if you're talking about evaluation caching, we would have to copy the evaluation cache sqlite databases which would result in an impure call to a file outside of the flake.

Perhaps we can lobby for a nix function that can evaluate an expression and output the evaluation cache.

blaggacao:
I think you are right. I'll test and report back. But probably you are right and my idea in https://github.com/Pacman99/devos/pull/2 is falling short of fixing this.

Pacman99:
I mean copying the drvPath is definitely an improvement. I assume it takes nix less time to get to the drv path than evaluate and get the store path.

blaggacao:
For lack of traction, for the te being I withdrawing from this issue.

PR #197, opened 2021-03-17
iso: only include the host's dependencies

### LINKED ISSUE #194: iso: systemd service startup
## Expected Behavior
The live cd acting as bespoke bootstrapping medium, only a necessary set of systemd modules should be started instead of starting up the complete os

## Current Behavior
While all relevant closures are included, and therefore act as a cache to the nixos installer, also all systemd services of the target host configuration are launched.

## Possible Solution
- Keep the closures & avoid service startup
- don't bother and just ssh into the target machine

## Steps to Reproduce
This happens with any current iso live installer.

## Context
Just a random example:
- A host is configured to launch cage and server a browser app in kiosk mode
- We could ssh into the target machine, here
- But still it might be beneficial to avoid service startup (there might be cases where this is indeed fully prohibitive, for example if those services involve some sort of state or host identity).

/cc @Pacman99

Pacman99:
fixes #194 

I've only tested that the iso build evaluates. I still have to test that it builds. And this change *could* break the current iso build, so it definitely has to be tested thoroughly. 

I came up with a pretty simple solution though, so I thought I may as well share it. Instead of importing all modules, I just added the hosts toplevel into the iso's storeContents.

cc @blaggacao

### DIFF
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -4,12 +4,13 @@
 lib.nixosSystem (args // {
   modules =
     let
+      moduleList = builtins.attrValues modules;
       modpath = "nixos/modules";
       cd = "installer/cd-dvd/installation-cd-minimal-new-kernel.nix";
 
       isoConfig = (lib.nixosSystem
         (args // {
-          modules = modules ++ [
+          modules = moduleList ++ [
             "${nixos}/${modpath}/${cd}"
             ({ config, ... }: {
               isoImage.isoBaseName = "nixos-" + config.networking.hostName;
@@ -58,7 +59,7 @@ lib.nixosSystem (args // {
           ];
         })).config;
     in
-    modules ++ [{
+    moduleList ++ [{
       system.build = {
         iso = isoConfig.system.build.isoImage;
       };
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -8,6 +8,8 @@ lib.nixosSystem (args // {
       modpath = "nixos/modules";
       cd = "installer/cd-dvd/installation-cd-minimal-new-kernel.nix";
 
+      hostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
+
       isoConfig = (lib.nixosSystem
         (args // {
           modules = moduleList ++ [
@@ -21,7 +23,10 @@ lib.nixosSystem (args // {
               nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
+                hostConfig.system.build.toplevel
+                hostConfig.system.build.toplevel.drvPath
               ];
+              environment.systemPackages = hostConfig.environment.systemPackages;
               # confilcts with networking.wireless which might be slightly
               # more useful on a stick
               networking.networkmanager.enable = lib.mkForce false;
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -10,11 +10,14 @@ lib.nixosSystem (args // {
 
       hostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
 
+      isoModules = (builtins.attrValues (builtins.removeAttrs modules [ "local" "lib" ]));
+
       isoConfig = (lib.nixosSystem
         (args // {
-          modules = moduleList ++ [
+          modules = isoModules ++ [
             "${nixos}/${modpath}/${cd}"
-            ({ config, ... }: {
+            ({ config, suites, ... }: {
+              imports = suites.base;
               isoImage.isoBaseName = "nixos-" + config.networking.hostName;
               isoImage.contents = [{
                 source = self;
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -25,9 +25,7 @@ lib.nixosSystem (args // {
               nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
-                self.devShell.${config.nixpkgs.system}.drvPath
                 hostConfig.system.build.toplevel
-                hostConfig.system.build.toplevel.drvPath
               ];
               environment.systemPackages = hostConfig.environment.systemPackages;
 
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -31,7 +31,6 @@ lib.nixosSystem (args // {
               ];
               environment.systemPackages = hostConfig.environment.systemPackages;
 
-              networking.hostName = hostConfig.networking.hostName;
               # confilcts with networking.wireless which might be slightly
               # more useful on a stick
               networking.networkmanager.enable = lib.mkForce false;
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -26,6 +26,7 @@ lib.nixosSystem (args // {
               nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
+                self.devShell.${config.nixpkgs.system}.drvPath
                 hostConfig.system.build.toplevel
                 hostConfig.system.build.toplevel.drvPath
               ];
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -10,14 +10,13 @@ lib.nixosSystem (args // {
 
       hostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
 
-      isoModules = (builtins.attrValues (builtins.removeAttrs modules [ "local" "lib" ]));
-
       isoConfig = (lib.nixosSystem
         (args // {
-          modules = isoModules ++ [
+          modules = moduleList ++ [
             "${nixos}/${modpath}/${cd}"
             ({ config, suites, ... }: {
-              imports = suites.base;
+              disabledModules = lib.remove modules.core suites.allProfiles;
+
               isoImage.isoBaseName = "nixos-" + config.networking.hostName;
               isoImage.contents = [{
                 source = self;
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -31,6 +31,8 @@ lib.nixosSystem (args // {
                 hostConfig.system.build.toplevel.drvPath
               ];
               environment.systemPackages = hostConfig.environment.systemPackages;
+
+              networking.hostName = hostConfig.networking.hostName;
               # confilcts with networking.wireless which might be slightly
               # more useful on a stick
               networking.networkmanager.enable = lib.mkForce false;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 
 **Documentation:**
 
+- doc: split iso [\#193](https://github.com/divnix/devos/issues/193)
 - lib: can depend on pkgs \(a la nixpkgs\#pkgs/pkgs-lib\) [\#147](https://github.com/divnix/devos/pull/147)
 
 **Closed issues:**
@@ -36,6 +37,7 @@
 
 **Merged pull requests:**
 
+- doc: enact bootstrapping section [\#196](https://github.com/divnix/devos/pull/196)
 - iso: copy input closourse into iso to avoide re-download [\#191](https://github.com/divnix/devos/pull/191)
 - add hosts module arg [\#164](https://github.com/divnix/devos/pull/164)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 
 **Merged pull requests:**
 
+- hosts/devosSystem: pass modules as attrset [\#198](https://github.com/divnix/devos/pull/198)
 - doc: enact bootstrapping section [\#196](https://github.com/divnix/devos/pull/196)
 - iso: copy input closourse into iso to avoide re-download [\#191](https://github.com/divnix/devos/pull/191)
 - add hosts module arg [\#164](https://github.com/divnix/devos/pull/164)
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -13,54 +13,46 @@ let
 
   suites = import ../suites { inherit lib; };
 
-  modules =
-    let
-      core = ../profiles/core;
-      modOverrides = { config, overrideModulesPath, ... }:
-        let
-          overrides = import ../overrides;
-          inherit (overrides) modules disabledModules;
-        in
-        {
-          disabledModules = modules ++ disabledModules;
-          imports = map
-            (path: "${overrideModulesPath}/${path}")
-            modules;
-        };
-
-      global = {
-        home-manager.useGlobalPkgs = true;
-        home-manager.useUserPackages = true;
-
-        hardware.enableRedistributableFirmware = lib.mkDefault true;
-
-        nix.nixPath = [
-          "nixpkgs=${nixos}"
-          "nixos-config=${self}/compat/nixos"
-          "home-manager=${home}"
-        ];
-
-        nixpkgs = { inherit pkgs; };
-
-        nix.registry = {
-          devos.flake = self;
-          nixos.flake = nixos;
-          override.flake = override;
-        };
-
-        system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+  modules = {
+    core = ../profiles/core;
+    modOverrides = { config, overrideModulesPath, ... }:
+      let
+        overrides = import ../overrides;
+        inherit (overrides) modules disabledModules;
+      in
+      {
+        disabledModules = modules ++ disabledModules;
+        imports = map
+          (path: "${overrideModulesPath}/${path}")
+          modules;
       };
 
-      # Everything in `./modules/list.nix`.
-      flakeModules =
-        builtins.attrValues self.nixosModules;
+    global = {
+      home-manager.useGlobalPkgs = true;
+      home-manager.useUserPackages = true;
 
-    in
-    flakeModules ++ [
-      core
-      global
-      modOverrides
-    ] ++ extern.modules;
+      hardware.enableRedistributableFirmware = lib.mkDefault true;
+
+      nix.nixPath = [
+        "nixpkgs=${nixos}"
+        "nixos-config=${self}/compat/nixos"
+        "home-manager=${home}"
+      ];
+
+      nixpkgs = { inherit pkgs; };
+
+      nix.registry = {
+        devos.flake = self;
+        nixos.flake = nixos;
+        override.flake = override;
+      };
+
+      system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+    };
+
+    # Everything in `./modules/list.nix`.
+    flakeModules = { imports = builtins.attrValues self.nixosModules ++ extern.modules; };
+  };
 
   specialArgs = extern.specialArgs // { inherit suites; };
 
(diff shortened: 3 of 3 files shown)


nrdxp:
This looks great! 

I was wanting to devise a way to actually test the iso as part of CI, but building the actual iso image itself is quite expensive both in space and time. Not sure if there would be any other way to reliably test it on each change though without having to build and squash the image.

If we can come up with something, perhaps we can include it in this PR.

blaggacao:
This is _really_ elegant, but at least we would need the (basic) user profiles so that remote access still works. Maybe re-importing the [base suite](https://github.com/divnix/devos/blob/core/suites/default.nix#L18) suffices. Basically replicate how it is done [here](https://github.com/divnix/devos/blob/bf9749b98a4a22135abc4cad525f81f42f597b0f/hosts/NixOS.nix#L4). That would solve this problem and not break the docs (though complementing them is necessary).

Pacman99:
> I was wanting to devise a way to actually test the iso as part of CI, but building the actual iso image itself is quite expensive both in space and time. Not sure if there would be any other way to reliably test it on each change though without having to build and squash the image.

I would like to see that too. But it does take a while to build the iso. Perhaps if we could just import the iso modules into the test node and inspect different parts of the machine once started up.

> Maybe re-importing the base suite suffices. Basically replicate how it is done here. That would solve this problem and not break the docs (though complementing them is necessary).

I didn't even think of that! Yeah I knew this would break things, thanks for figuring what exactly it broke :) 
So I tried adding the base suite, but then that runs into the problem of home-manager missing. I think getting that to work will require a rework of `hosts/default.nix` and `devosSystem`. We might want to consider not passing ALL the modules as one list and instead seperating them. Or passing them as an attrset, so different parts of devosSystem can pick apart the necessary modules. And then a final `attrValues` of the attrset could be used in the final host.

Pacman99:
With #198, I can selectively import the base suite and flakeModules - which contains home-manager. Since those are the only two that declare new options(I think!), I think that should get any user's base suite to evaluate.

blaggacao:
I think `core` is valuable too for the unified "sysadmin experience".

Pacman99:
Yeah I'm actually thinking now of just including everything but local and lib.

blaggacao:
That is a good idea, the environments should be probably as close as possible in every aspect, just not starting more systemd services than declared in `suites.base` and `profiles/core`.

I have a hard time following up on the home-mamager problwm you reported. What is it about?

Pacman99:
> That is a good idea, the environments should be probably as close as possible in every aspect, just not starting more systemd services than declared in suites.base and profiles/core.

I add a change to forward environment.systemPackages. But either way every other dependency is in the store, they just have to be installed to the profile.

> I have a hard time following up on the home-mamager problwm you reported. What is it about?

So the base suite imports the users which set the home-manager options. But without importing extern, those options don't exist, so the evaluation fails. And it indicates a general problem with trying to import parts of the host's config without the necessary modules. If a user had declared options in a custom module, then importing the base suite would then fail.

blaggacao:
Ok, I understand. So we need to make sure to actually import all _modules_ but only core, nixos-user and root-user profiles. Right?

blaggacao:
For a live installer we actually only need core + root. We can drop nixos. We should actually drop nixos, since deploy-rs does require _passwordless_ sudo and it's quite easy to set a password for `nixos` during install (you are prompted).

Pacman99:
Filtering out users is a good idea, but does the root user have ssh access to the iso by default? So if I understand right, the user adds an ssh public key to the root user module which is imported to the iso build. And then you can now ssh into the system as the root user.  But we don't want to include the other users as that would be unnecessary and will likely have a password.

So instead of importing the base suite, I'm now just importing `suites.allUsers.root`.

nrdxp:
I have occasion to test this out as it looks like I fudged my zfs ashift settings and need to recreate my pool. I'll report back how it goes.

Pacman99:
Updated to include core, global, flakeModules, and modOverrides. The last two aren't really necessary, but they will prevent an evaluation failure in the rare case someone uses home-manager or sets some other option expecting an overridden module

blaggacao:
> but does the root user have ssh access to the iso by default?

No, but it is documented how to do that. Wording might be in need of alight tweaking after this discussion.

lib/devos/devosSystem.nix
```diff
@@ -8,10 +8,17 @@ lib.nixosSystem (args // {
       modpath = "nixos/modules";
       cd = "installer/cd-dvd/installation-cd-minimal-new-kernel.nix";
 
+      hostConfig = (lib.nixosSystem args).config;
```
nrdxp, line 11: This needs to be `hostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config` due to modules being a set now.

blaggacao:
> Updated to include core, global, flakeModules, and modOverrides.

I think this is actually necesarry (and all other _modules_, too) in order to be robust.

This is why I would rather invert the logic woth a selector:

- everytjing
- less all suites.allProfiles - except `core`
- less all suites.allUsers - except `root`

Such a negative selector is much more robust than a positive one. Actually a positive one is not correct here.

lib/devos/devosSystem.nix
```diff
@@ -8,10 +8,17 @@ lib.nixosSystem (args // {
...
       isoConfig = (lib.nixosSystem
         (args // {
-          modules = moduleList ++ [
+          modules = [
             "${nixos}/${modpath}/${cd}"
+            modules.core
+            modules.modOverrides
+            modules.flakeModules
+            modules.global
+            hostConfig.lib.specialArgs.suites.allUsers.root
```
nrdxp, line 21: This doesn't work as `allUsers` is actually a list not a set. We can just export users in suites, which we currently don't do, but assuming users haven't modified this may be a bad idea, as I actually deleted this user on my personal config and it caused a failure. Plus the default root user only sets an empty root password anyway, which I think the ISO config already does.

Pacman99: I think, at least for now, I'm just going to import the base suite. So that way all users are imported and it gives the ability for users to configure whats included in the iso.

Pacman99: I think the main goal here is to allow users to set their ssh public key.

lib/devos/devosSystem.nix
Pacman99: fixed

lib/devos/devosSystem.nix
```diff
@@ -8,10 +8,16 @@ lib.nixosSystem (args // {
...
+
       isoConfig = (lib.nixosSystem
         (args // {
-          modules = moduleList ++ [
+          modules = hostConfig.lib.specialArgs.suites.base ++ [
             "${nixos}/${modpath}/${cd}"
+            modules.core
+            modules.modOverrides
+            modules.flakeModules
+            modules.global
```
blaggacao, line 20: How does future self know that it needs to add lines here if it modifies hosts/default.nix adding more modules?

blaggacao (changes requested):
This is the reason for why I'm more in favor of anegative selector ...

Pacman99:
@blaggacao I'm not entirely sure what you mean? Isn't the goal here to only import the necessary things as to prevent unecessary system startup while still copying the closure. And with that in mind I would think a positive selector is better. So in the future if more modules get added, we have to go through and evaluate if they should be in the isoConfig.

Could you perhaps share a nix expression as to what should be imported. I think might clear up confusion.

lib/devos/devosSystem.nix
Pacman99: I feel like thats a good thing, that we have to evaluate the importance of modules added to hosts/default.nix before adding them to isoConfig.

nrdxp:
If we really wanted to go deep. We could pass args to each module manually, and then, if the module happens to define a systemd service, we can filter it out of it's config session before including it. Might be more complex than it sounds though.

lib/devos/devosSystem.nix
Pacman99: >Future self can silently break things beyond the reach of the hosts/default.nix scope.

I think theres potential to break things either way. moving the discussion for positive vs. negative selector here.

blaggacao:
The goal can be reached in two ways, indeed.

Per #194 we want ro avoid systemd to start up. If I recall the docs correctly, those type of things are _only_ defined in profiles. That might be user profiles (in theory: user systemd serbices) or regular profiles.

No systemd serbice is ever avtivated by a _module_ (maybe specified, etc. But not switched on).

But modules define options that a profile (which we need here) might use. We just don't know if and which one.

Hence, we need to include _all_ modules, but no profile, except the ones we clearly document,so that users will know, that if they put a systemd service in one of those profiles, it will start as part of the live installer.

I suggest, we only document `core` and `root`. (And if there is no root, nothing bad happens, you just can't ssh into the live installer, where the only user you really need is root).

blaggacao:
So the logic would be: +-

`modulesList - suites.allProfiles - suites.allUsers + core + root`

blaggacao:
If beeing able to ssh into root is unwanted, we can also choose nixos insstead. I think that might be a concern, indeed. We just would need to make sure that for `deploy-rs` this user has _passwordless_ sudo. But maybe that can actually be considered an issue in upstream deploy-rs.

So let's rather use `nixos` and file an issue in deploy-rs.

Pacman99:
Okay I think I clearly follow what your saying and I agree. I just pushed a commit that I believe accomplishes it. Its all modules except local and lib. Which is essentially `modulesList - allProfiles`. And as I said earlier, I'm just importing the entire base suite. I don't know how to filter out users, and also I don't see a point in doing so. User services are usually really light if any are even defined. And it could be nice to login to your user and get an environment your comfortable with.

blaggacao:
Agreed! I'll file an issue in deploy-rs then.

Pacman99:
I think there is an issue in deploy-rs for password based sudo. Unless I misunderstand and your talking about something else.

blaggacao:
This is starting to look pretty clean. i can test tomorrow on my setup end to emd manually for a final round pf feedback / approval. 👍

nrdxp:
> I'm in an excellent position to try this out since I'm deploying a cage kiosk service, and before this PR I would have to manually build with cage switched to opening a terminal instead of a browser.

After testing tomorrow, if you could compile what you learn about bootstrapping and write a nixos integration test for the iso config, that would be amazing. I especially want to test the link local stuff to ensure it's actually working as we move forward.

lib/devos/devosSystem.nix
```diff
@@ -21,7 +26,10 @@ lib.nixosSystem (args // {
               nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
```
blaggacao, line 28: We should be able to add `drvPath` for devshell, as well, fixing #195, shall we?

Not meaning to block this PR, though.

blaggacao: x-ref: https://github.com/Pacman99/devos/pull/2 (also going to test that)

Pacman99: good idea, merged!, But will that really cache evaluations? I thought evaluation caches were sqlite databases in your home directory. I remember it being a series of sqlite db's with hashe's as names

nrdxp: I believe @Pacman99 is right about the evaluation cache. I've been struggling with how to clear the dang thing for a while now. Pretty annoying when the error message for a failed derivation just becomes the generic:
```console
failure of cached attribute
```
I know I found it's exact location once in order to clear it, but now I can't seem to find it. If it is still in the home dir, perhaps we could make some clever use of home-manager to include it.

Pacman99: Ohh for me its in `~/.cache/nix/eval-cache-v2`

Pacman99: you should be able to safely delete that since its just evaluations that can be done again

Pacman99:
Just force pushed with a cleanup of the git history. Its just three commits now.

bors try

blaggacao:
It looks like [this](https://github.com/NixOS/nixpkgs/blob/80c7312da684a0713464313a744486b3b8d66587/nixos/modules/installer/cd-dvd/iso-image.nix#L427-L438) is now taking _forever_. I'll try with `zstd`, but I suspect we rather need to generate an overlay fs (if possible) and store those additional closures (not necessary at boot) just plainly on the files system, without compression and without inclusion into the squashfs.

blaggacao:
bors try-

lib/devos/devosSystem.nix
```diff
@@ -8,11 +8,16 @@ lib.nixosSystem (args // {
       modpath = "nixos/modules";
       cd = "installer/cd-dvd/installation-cd-minimal-new-kernel.nix";
 
+      hostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
+
+      isoModules = (builtins.attrValues (builtins.removeAttrs modules [ "local" "lib" ]));
```
blaggacao, line 13: Result of my tests: removing `local` removes the correct hostname. The result is a badly named iso image:

`result/iso/nixos-nixos-20.09.20210312.60b18a0-x86_64-linux.iso` instead of

`result/iso/nixos-POS-20.09.20210312.60b18a0-x86_64-linux.iso`

blaggacao (changes requested):
I'll make another try re-adding `local`. Maybe we can later think about tearing local into smaller pieces in search of a concise interface. This breaks image naming...

lib/devos/devosSystem.nix
Pacman99: I can pass hostName manually

blaggacao:
I think we should revert to strictly solve #194 and leave the other stuff for another PR. It is too hard to isolate and understand cause-effects relationship with mixed concerns if every end-to-end test takes more than 2 hours.

Pacman99:
> I think we should revert to strictly solve #194 and leave the other stuff for another PR.

Isn't that the entire point of all the changes in this PR? I'm confused as to what you would like to revert to. Either we import all modules or figure out a way to include the closure and import necessary modules. This PR is trying to do the latter.

> It is too hard to isolate and understand cause-effects relationship with mixed concerns if every end-to-end test takes more than 2 hours.

I think its ok if we take some time to figure out this PR. Also another way to test could be to build the isoConfig's toplevel and inspect the files in there.

blaggacao:
I'm gonging to test now alternative implementation https://github.com/divnix/devos/pull/202 which tries not to introduce side effects. It still includes https://github.com/divnix/devos/pull/202/commits/2bd331429d7a89f0f94aeaf36fff6195f25cd240 as we already started to go down that road with https://github.com/divnix/devos/pull/198 although its not strictly necessary for solving #194

Pacman99:
Borrowed @blaggacao's idea of using disabledModules here. And I think this should avoid potential side effects.

lib/devos/devosSystem.nix
```diff
@@ -21,7 +25,12 @@ lib.nixosSystem (args // {
               nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
+                self.devShell.${config.nixpkgs.system}.drvPath
+                hostConfig.system.build.toplevel
+                hostConfig.system.build.toplevel.drvPath
               ];
+              environment.systemPackages = hostConfig.environment.systemPackages;
```
blaggacao, line 32: What was the rationale (use case) for this again?

Pacman99: Well by including the hostConfig toplevel, all packages are now in the nix store. So I thought as  convenience we could just forward to the environment.

blaggacao:
I get a segfault on its current state, maybe the `lib.remove` doesn't work as intended?  #202 builds now an I could start into the live installer, going to rebase my cage project on #202 to finally verify #194.
```
/nix/store/vf2f1xph3v6afy0bx0cf46bp82qincpy-devshell-dir/bin/flk: line 28: 308429 Segmentation fault      (core dumped) nix build "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" "${@:3}
```

Pacman99:
`nix build .#nixosConfigurations.NixOS.config.system.build.iso` works for me Not sure where the segfault comes from. But that seems like a nix issue.

lib/devos/devosSystem.nix
blaggacao: I'm actually not sure if we can include the toplevel derivation, it takes a lot more time to produce the squashfs, now. Well, maybe we can leave it, but open an issue in search for a better solution (eg. wo compression?).

So we might just leave this line, as well.

Pacman99: I'm currently attempting to build and I see what you mean. If this PR increases the time to build or the iso size by a lot, it might not be the best solution.

Perhaps we could just forward the systemPackages and not include the toplevel build. Or we could add a separate build that has the entire toplevel. So one for network install and one that includes everything making an offline install possible.

lib/devos/devosSystem.nix
```diff
@@ -20,7 +25,12 @@ lib.nixosSystem (args // {
               nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
+                self.devShell.${config.nixpkgs.system}.drvPath
+                hostConfig.system.build.toplevel
```
blaggacao, line 29: It looks like this is actually redundant: https://github.com/NixOS/nixpkgs/blob/a6155fc19e31fe0a49b75409d02a2d8b6bb97aa0/nixos/modules/installer/cd-dvd/iso-image.nix#L627-L638

and we need to remove the `*.drvPath`. THose are the size culprits

Pacman99: Ohh I see, fixed in latest commit

Pacman99: what does includeSystemBuildDependencies mean? I would assume if you pass the toplevel, that would include all buildInputs for that store path.

blaggacao:
Can you also merge https://github.com/Pacman99/devos/pull/3? So I can try again off of this branch here?

lib/devos/devosSystem.nix
blaggacao: You can also remove the `.toplevel`, it's already included as in the linked default config is shown.

> what does includeSystemBuildDependencies mean?

https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/cd-dvd/iso-image.nix#L482-L486

blaggacao:
@Pacman99 For various reasons, this has taken me a whole day already, and I still wasn't able to test this fully. Are you ok, if we turn over the page and continue on #202 ? It's just too hard for me to keep track of all those things while not being able to commit here... (sry)

lib/devos/devosSystem.nix
Pacman99: are you talking about this line: https://github.com/NixOS/nixpkgs/blob/a6155fc19e31fe0a49b75409d02a2d8b6bb97aa0/nixos/modules/installer/cd-dvd/iso-image.nix#L630?

That would get the toplevel of the isoConfig. Which won't include the profiles that have been disabled. And the goal is to include the profile's closure, just without including them in the config.

Pacman99:
I'll leave this PR open as a draft. But yeah this was mainly meant to demonstrate an idea and I don't think that the idea is viable. I think we'll have to experiment more to get a better overall solution. But I'm not sure #202 is much better since its basically the same as this PR but with a few tweaks.

lib/devos/devosSystem.nix
blaggacao: Yep, just realized that my self and renamed to `fullHostConfig` and left a [comment](https://github.com/blaggacao/devos/blob/b9a256605d6b7e77e787c0565b8e42bb7427a1f9/lib/devos/devosSystem.nix#L32-L34). That makes the difference clearer.

Pacman99:
Perfectly fine with continuing there though since you're the one doing most of the testing anyways itll be easier for you to update.

Pacman99:
I think the CI runner is still trying to evaluate this PR, but I don't think the CI runner is even available right now.

Pacman99:
Included in #202

PR #198, opened 2021-03-17
hosts/devosSystem: pass modules as attrset

Pacman99:
This is a fairly simple change that only changes the lib api for devosSystem. But doesn't add any features by itself. Hosts now pass modules to devosSystem as an attrset. And devosSystem just grabs all modules in the set and passes it to nixosSystem. 

I plan to use this in #197 to selectively import modules. And I think it could help with nix-darwin - and other config systems - support, since not all profiles and modules are config system agnostic. This could be a workaround to add rudimentary support for other config systems by only importing the necessary modules.

Overall I think its a useful change and extends the abilities of `devosSystem`

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -13,54 +13,46 @@ let
 
   suites = import ../suites { inherit lib; };
 
-  modules =
-    let
-      core = ../profiles/core;
-      modOverrides = { config, overrideModulesPath, ... }:
-        let
-          overrides = import ../overrides;
-          inherit (overrides) modules disabledModules;
-        in
-        {
-          disabledModules = modules ++ disabledModules;
-          imports = map
-            (path: "${overrideModulesPath}/${path}")
-            modules;
-        };
-
-      global = {
-        home-manager.useGlobalPkgs = true;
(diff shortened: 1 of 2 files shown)


hosts/default.nix
```diff
@@ -13,85 +13,77 @@ let
...
-          "${toString ./.}/${hostName}.nix"
-        ];
-
-        networking = { inherit hostName; };
-
-        _module.args = {
-          inherit self;
-          hosts = builtins.mapAttrs (_: host: host.config)
-            (removeAttrs hosts [ hostName ]);
+      localModules = {
```
blaggacao, line 61: I stumbled over this semantic. I think it is cleare to not pack them and just do blow:

```nix
{
   modules = modules // { inherit lib local; };
}
```

blaggacao (changes requested):
Slight semantical ironing out...

hosts/default.nix
Pacman99: Yeah that looks nicer

blaggacao (approved):
LGTM

Pacman99:
bors try

Issue #200, opened 2021-03-18
potential naming clash


blaggacao:
https://github.com/divnix/devos/blob/2961c1ea529a98296be9c0f07a08431665bb8654/lib/devos/mkProfileAttrs.nix#L30

this attribute could be called `profile` instead, since this naming scheme precludes people naming their **sub** profile `default` &mdash; which people might actually want to do.

Is my analysis correct, here?

blaggacao:
I think I'm completely confused :confused:

blaggacao:
> since this naming scheme precludes people naming their **sub** profile default — which people might actually want to do.

I realized that this restriction only applies for sub profiles, a top level profile still can be named `default`, as I understand things.

naming a sub level profile default alongside a `default.nix` is probably not going to happen:

```
core/
core/default.nix
core/default/default.nix
```

... improbable.

PR #201, opened 2021-03-18
suites: simplify code

blaggacao:
If this code had been written this way, I wouldn't have opened #200 in
the first place. (perspective of a begginer nix-langer)

closes: #200 This is a pre-fix to PR #202

### DIFF
--- a/lib/devos/mkProfileAttrs.nix
+++ b/lib/devos/mkProfileAttrs.nix
@@ -11,6 +11,19 @@ let mkProfileAttrs =
   Example:
   let profiles = mkProfileAttrs ./profiles; in
   assert profiles ? core.default; 0
+
+  Example folder layout:
+  profiles/core/default.nix
+  profiles/core/sub/default.nix
+  profiles/code/ignored-no-default-nix/
+  profiles/other/default.nix
+
+  Example result:
+  {
+    core = { default = "..." };
+    sub = { default = "..." };
+    other = { default = "..." };
+  }
   **/
   dir:
   let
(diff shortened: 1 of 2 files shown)


Pacman99:
This makes defining suites much more complex. You have to call profileMap at the front of every suite.

blaggacao:
> This makes defining suites much more complex. You have to call profileMap at the front of every suite.

But at least I do understand what I'm actually doing :wink:

blaggacao:
bors try+

blaggacao:
bors try

blaggacao:
bors try-

I think I still don't understand what this code does.

Pacman99:
>profileMap implements the receiving end of the fs layout contract. Now everything is clear by analogy / parallelism with allProfiles & allUsers.

I think proper ux and a stable api is more important than an analogy being correct. And having users run a function on every suite manually, imo, is a step backwards. And if you just look down at the mapAttrs, you would see that profileMap is run at every function. So I really don't see what this PR accomplishes.

blaggacao:
ping, made a change.

The commit message has the impetus (me in abstractable role of nix beginner trying to understand what's going on).

blaggacao:
bors try

blaggacao:
This code produces a `segfault`. Whooo :waffle: 

```
/nix/store/vf2f1xph3v6afy0bx0cf46bp82qincpy-devshell-dir/bin/flk: line 28: 300128 Segmentation fault      (core dumped) nix build "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" "${@:3}
```

I was thinking it would have been equivalent.

PR #202, opened 2021-03-18
iso: avoid systemd service startup

### LINKED ISSUE #194: iso: systemd service startup

blaggacao:
fixes #194
alternative to #197

# Manual Tests

<details>
<summary>was unrelated</summary>

- [ ] `flk install NixOS --impure` correctly onto `/mnt` :negative_squared_cross_mark: (looks like no profile is present)

Issue: https://github.com/divnix/devos/issues/204
Upstream Issue: https://github.com/NixOS/nixpkgs/issues/116938

</details>

- [x] acceptable build time / closure size ca 850MB (for a simple base OS) :heavy_check_mark: 
- [x] local profile with `cage` service is disabled, that is: boots into terminal :heavy_check_mark: 
- [x] success: air gapped / offline devshell enter :heavy_check_mark: 
- [ ] failure: aire gapped target install: &larr; non blocking bonus item :negative_squared_cross_mark: 
```console
$ flk install POS
warning: you don't have internet access; disabling some network-dependent features
building the flake in path:/iso/devos?narHash=sha265-...
warning: you don't have internet access; disabling some network-dependent features
error: unable to download 'https://api.github.com/repos/NixOS/nixpkgs/df8e3...': Couldn't resolve host name (6)
```

&rarr; detailed rationale in the commit messages

:heart: @Pacman99 for the excellent and detailed discussions in #197 and the may ideas, suggestions and code.

### DIFF
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -7,6 +7,8 @@ lib.nixosSystem (args // {
       moduleList = builtins.attrValues modules;
       modpath = "nixos/modules";
 
+      fullHostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
+
       isoConfig = (lib.nixosSystem
         (args // {
           modules = moduleList ++ [
@@ -27,6 +29,9 @@ lib.nixosSystem (args // {
               }];
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
+                # include also closures that are "switched off" by the
+                # above profile filter on the local config attribute
+                fullHostConfig.system.build.toplevel
               ];
 
               # confilcts with networking.wireless which might be slightly
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -33,6 +33,8 @@ lib.nixosSystem (args // {
                 # above profile filter on the local config attribute
                 fullHostConfig.system.build.toplevel
               ];
+              # still pull in tools of deactivated profiles
+              environment.systemPackages = fullHostConfig.environment.systemPackages;
 
               # confilcts with networking.wireless which might be slightly
               # more useful on a stick
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -12,7 +12,11 @@ lib.nixosSystem (args // {
         (args // {
           modules = moduleList ++ [
             "${nixos}/${modpath}/${cd}"
-            ({ config, ... }: {
+            ({ config, suites, ... }: {
+
+              # avoid unwanted systemd service startups
+              disabledModules = lib.remove modules.core suites.allProfiles;
+
               isoImage.isoBaseName = "nixos-" + config.networking.hostName;
               isoImage.contents = [{
                 source = self;
--- a/lib/devos/devosSystem.nix
+++ b/lib/devos/devosSystem.nix
@@ -6,31 +6,35 @@ lib.nixosSystem (args // {
     let
       moduleList = builtins.attrValues modules;
       modpath = "nixos/modules";
-      cd = "installer/cd-dvd/installation-cd-minimal-new-kernel.nix";
 
       isoConfig = (lib.nixosSystem
         (args // {
           modules = moduleList ++ [
-            "${nixos}/${modpath}/${cd}"
+
+            "${nixos}/${modpath}/installer/cd-dvd/installation-cd-minimal-new-kernel.nix"
+
             ({ config, suites, ... }: {
 
               # avoid unwanted systemd service startups
               disabledModules = lib.remove modules.core suites.allProfiles;
+
+              nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
 
               isoImage.isoBaseName = "nixos-" + config.networking.hostName;
               isoImage.contents = [{
                 source = self;
                 target = "/devos/";
               }];
-              nix.registry = lib.mapAttrs (n: v: { flake = v; }) inputs;
               isoImage.storeContents = [
                 self.devShell.${config.nixpkgs.system}
               ];
+
               # confilcts with networking.wireless which might be slightly
               # more useful on a stick
               networking.networkmanager.enable = lib.mkForce false;
               # confilcts with networking.wireless
               networking.wireless.iwd.enable = lib.mkForce false;
+
               # Set up a link-local boostrap network
               # See also: https://github.com/NixOS/nixpkgs/issues/75515#issuecomment-571661659
               networking.usePredictableInterfaceNames = lib.mkForce true; # so prefix matching works
--- a/lib/devos/mkProfileAttrs.nix
+++ b/lib/devos/mkProfileAttrs.nix
@@ -27,7 +27,7 @@ let mkProfileAttrs =
     f = n: _:
       lib.optionalAttrs
         (lib.pathExists "${dir}/${n}/default.nix")
-        { default = "${dir}/${n}"; }
+        { default = /. + "${dir}/${n}"; }
       // mkProfileAttrs "${dir}/${n}";
   in
   lib.mapAttrs f imports;


lib/devos/devosSystem.nix
```diff
@@ -6,27 +6,38 @@ lib.nixosSystem (args // {
...
           modules = moduleList ++ [
-            "${nixos}/${modpath}/${cd}"
-            ({ config, ... }: {
+
+            "${nixos}/${modpath}/installer/cd-dvd/installation-cd-minimal-new-kernel.nix"
+
+            ({ config, suites, ... }: {
+
+              # avoid unwanted systemd service startups
+              disabledModules = suites.allProfilesButCore;
```
Pacman99, line 21: This is a good idea, I didn't even know about diabledModules. Although looking at the nixpkgs manual, are you sure they aren't supposed to be paths from nixpkkgs?

Instead of modifying suites, this can just be done as `lib.remove modules.core suites.allProfiles`

blaggacao: > Although looking at the nixpkgs manual, are you sure they aren't supposed to be paths from nixpkkgs?

I've to clear the noodle soup in my head through further testing. :wink: 

> Instead of modifying suites, this can just be done as lib.remove modules.core suites.allProfiles

Way better, thanks! (noodle soup, there is!)

blaggacao: > Way better, thanks! (noodle soup, there is!)

Actually that doesn't work it was caused by a mishap of mine in #200 (force-pushed fixed, now)

blaggacao:
```
/nix/store/vf2f1xph3v6afy0bx0cf46bp82qincpy-devshell-dir/bin/flk: line 28: 289330 Segmentation fault      (core dumped) nix build "$DEVSHELL_ROOT#nixosConfigurations.$2.config.system.build.iso" "${@:3}"
```

:question: :question: :question:

blaggacao:
bors try

lib/devos/devosSystem.nix
```diff
@@ -6,27 +6,42 @@ lib.nixosSystem (args // {
...
           modules = moduleList ++ [
-            "${nixos}/${modpath}/${cd}"
-            ({ config, ... }: {
+
+            "${nixos}/${modpath}/installer/cd-dvd/installation-cd-minimal-new-kernel.nix"
+
+            ({ config, suites, ... }: {
+
+              # avoid unwanted systemd service startups
+              disabledModules = lib.remove modules.core suites.allProfiles;
```
Pacman99, line 21: I know I suggested this change, but the fact that `mkProfileAttrs` has to be edited for it to work indicates that its somewhat finicky. If the format of modules.core ever changes or of allProfiles. This will stop the core profile from being imported.

Pacman99: maybe you could instead do `builtins.filter (x: import modules.core != import x) suites.allProfiles`

Pacman99:
Also apparently getting a segmentation faullt while building the iso is a known issue with nix that happens sometimes. A workaround is to preface the build command with `GC_DONT_GC=1`

lib/devos/devosSystem.nix
blaggacao: As per the second part of the respective commit message, this is a hard requirement for disabledModules to work at all as we expect it to work.

<details>
<summary>Stance</summary>

Besidws of that, I also think passing around the _correct_ type (paths) is a contract that the caller should expect from an interface.

I know nix-lang is not precicely the best language in declaring and enforcing contracts.

Fortunately, Nickel will solve that problem. Until then, we are left with something like crossing fingers and being carful. 😉 

</details>

blaggacao:
bors try

lib/devos/devosSystem.nix
Pacman99: Ohh yeah I'm okay with changing lib too as part of this pr. Passing profiles around as a path is definitely a good idea. But `lib.remove` is a bit finicky with modules since they can be in muliple formats. So I think directly using `builtins.filter` is much more consistent.

blaggacao:
I experience that `nixos-install --flake ..` (that is `flks install NAME`) is equivalent to `nixos-install` based upon a generated `/mnt/etc/nixos/configuration.nix`.

Is `nixos-install` currently broken?

Once I delete `/mnt/etc/nixos/configuration.nix`, it says:

configuration file `/mnt/etc/nixos/configuration.nix`q dosen't exist.

Can `nixos-install` only be executed on a freshly formatted disk and has unexpected behavior otherwise?

blaggacao:
Indeed it is:
https://github.com/NixOS/nixpkgs/blob/e30d87f22237fe25618136eb9f818ecefaeb02a9/nixos/modules/installer/tools/nixos-install.sh#L128
in combination with: https://github.com/NixOS/nix/issues/4613 / https://github.com/NixOS/nix/commit/66fa1c7375e4b3073a16df4678cf1d37446ed20b

lib/devos/devosSystem.nix
Pacman99: Can you switch this line to use `builtins.filter` and import the modules before comparing. Even with the change you made in lib, comparing two imported modules is more consistent than comparing paths.

Pacman99 (approved):
Looks great

Pacman99:
> failure: aire gapped target install: ← non blocking bonus item

Just wondering, is this because of not importing `drvPath`? Perhaps you could add another build for an offline install iso vs a network install iso. But as you said thats a bonus

blaggacao:
> Just wondering, is this because of not importing `drvPath`?

If I remember correctly, it was trying to download `nixpkgs` source. &mdash; which made me wonder about the registry pinning. But I think to fully understand, we would have to dig through https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-install.sh

Maybe we can open an issue for air-gapped install, and try to make it work over time.

lib/devos/devosSystem.nix
blaggacao: > NVM I think comparing paths is correct

Yeah, string can be just arbitrary relative paths, unlink paths, they have no implicit guarantees to be canonical.

Pacman99:
I think it might try and download the registry anyways and perhaps try to update flake. I think theres a `--offline` flag that you could try and see how far you get.

lib/devos/mkProfileAttrs.nix
```diff
@@ -27,7 +27,7 @@ let mkProfileAttrs =
     f = n: _:
       lib.optionalAttrs
         (lib.pathExists "${dir}/${n}/default.nix")
-        { default = "${dir}/${n}"; }
+        { default = /. + "${dir}/${n}"; }
```
nrdxp, line 30: What's the motivation here?

Pacman99: for the expression in `devosSystem` `lib.remove modules.core suite.allProfiles`. For the matching to work all items in the list must be the same type. And in general its better to pass around modules as paths rather than strings.

lib/devos/devosSystem.nix
```diff
@@ -6,27 +6,42 @@ lib.nixosSystem (args // {
     let
       moduleList = builtins.attrValues modules;
       modpath = "nixos/modules";
-      cd = "installer/cd-dvd/installation-cd-minimal-new-kernel.nix";
+
+      fullHostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
 
       isoConfig = (lib.nixosSystem
```
nrdxp, line 12: I would really love to see the bonus goal of your PR implemented. Is it possible that we could create an `isoConfig'` that is everything that the current `isoConfig` defines with `storeContents` abstracted out into a separate variable, while the new `isoConfig` would just be something like:
```nix
{
  isoConfig = isoConfig' // {  isoImage.storeContents = storeContents ++ [ isoConfig' ];
}
```

I'm trying to think of a way to do this without triggering infinite recursion, although perhaps I haven't thought this through enough and the above does cause and infinite loop. I'll have more time to test tomorrow and see if I can work it out.

lib/devos/mkProfileAttrs.nix
blaggacao: But also.most importantly for `disabledModules` to filter.on `module.key` proper instead of a rleative to.`modulePath` string.

lib/devos/devosSystem.nix
blaggacao: Within the `devshell` of a live iso, it is ultimately [upstream nixos-install at this line](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-install.sh#L128) that is hindering us, I guess: `nix flake metadata` always wants to evaluate a lot. I dodn't even understand why this line js there. It returns the store-path lf the derivation. Couldn't we obtain that more cheaply?


The [build](github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-install.sh#L156) itself is already fully cached (`auto?trusted=1`) courtesy of the `fullHostConfig`'s `toplevel`. That extends tl the system profile, which is [loaded locally from the build's output](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/nixos-install.sh#L167)

After that there is no potential network call left. So the culprit must be `nix flake metadata`...

nrdxp: If it is only nixos-install that's blocking us (and it has other [problems](https://github.com/NixOS/nix/issues/4081#issuecomment-753237142) with flakes), perhaps we could just do what's mentioned in that comment, or something similar to work around it.

blaggacao: We just moved to a new appartment and, for a varying operational reasons, I probably will be unavailable to follow trough on this bonus item any time soon. (No inet, no setup, obly cell-phone, life getting in the way, ...)

Can we carve this out into an issue while moving on with this PR in its current state?

nrdxp: Sure. I'd like to have a crack at it first though, if you don't mind, as I think it would be pretty straight-forward. If I get it working, I'll pushing a PR to your branch.

blaggacao: I suspect the `--offline` option would get us quite some mileage (combined with a small patch to `nixos-install` / flakes-first reconception of it), unfortunately I havn't been able to test this in time.

nrdxp: seems I don't have as much time as I'd hoped. I'll go ahead and merge this and we will try to work it out later.
bors r+

Issue #91, opened 2021-01-12
flk sync (with mothership)


blaggacao:
**Is your feature request related to a problem? Please describe.**
Once upon a time, well into `new flake -t nrdxp/nixflk`, user wants to rejuvenate its repo state with the latest and greatest additions of the mother ship (`nrdxp/nixflk`).

Not having done a proper clone, but using the "blessed" flake facilities, user needs to clasical-patch his source tree on unrelated histories.

**Describe the solution you'd like**
`flk sync` that does `git format-patch` on the mothership since the time where `nix flake new` was invoked. It also helps applying them and hands off to git tooling where merge conflicts apply.

**Describe alternatives you've considered**
- Not everybody is familiar with the alternative forking workflow, especially since `nix flake` has the "blessed" way of using templates.
- This kind of workflow could alternatively be supported by flakes directly (maybe out of scope? - at least short to mid term)

Once initiated a template travking the mothership manually is daunting, error-prone and in-efficient. Patching at least mitigates the latter two. Conflict resolution might remain daunting.

**Additional context**
None.

blaggacao:
x-ref: https://github.com/NixOS/nix/issues/4474

blaggacao:
Directly after commiting the templated files something like this could be done:
```
git notes add -m "https://github.com/nrdxp/nixflk/compare/24eeb3e02a84e332ce68b80ab9db53caab244ce0...bare.patch"
```

then we can access pull in updates like so:
```console
$ curl $(echo "$(git notes show $(git notes list | tail -n 1 | cut -d' ' -f2))") | git apply  # might be `head` instead
$ git add .
$ git commit -m "sync with mothership"
$ git notes add -m "https://github.com/nrdxp/nixflk/compare/$(TODO: get new remote commit?)...bare.patch"
```

Damn, is this brittle, but it looks nice in `git log`:
```console
commit 7ff7ca54c21df18423be6eb09eae8b7951908c1e (HEAD -> tmp-main)
Author: David Arnold <dar@xoe.solutions>
Date:   Mon Jan 25 15:47:43 2021 -0500

    Add bare ./Tech (nrdxp/nixflk)

Notes:
    https://github.com/nrdxp/nixflk/compare/24eeb3e02a84e332ce68b80ab9db53caab244ce0...bare.patch
```

PR #206, opened 2021-03-20
mkDevos and evalDevosArgs: Move flake implementation logic into lib

Pacman99:
Moving all logic to lib makes things easier to change and understand, since whenever you add a feature, you only have to edit relevant functions in lib. And for a user, they don't have to understand whats happening behind the scenes. But if they want to I think reading functions in lib is simpler than reading implementation logic thats in different places around the directory.

But most importantly this allows for a lib function called `mkDevos` which is a potential solution to #91. The mkDevos function can take other an input of other flakes and build a devos flake just like this repo does. So this creates an alternate method of using devos, where you just add devos as an input then do
```
devos.lib.mkDevos { inherit self; }
```
And then you can just create the relevant directories(this could be shipped as its own template) and those directories will be imported. This change is also backwards compatible as demonstrated in the updated flake.nix where the current devos just calls `mkDevos` with itself. So a user can either just clone this repo and use it like its currently being used, or they can use the lib function.. The advantage of just using the lib function is you no longer have to worry about merging upstream and conflicts that arise, you can just update your devos input, so nix flake takes care of syncing with the mothership for you.

Keep in mind you have to add all inputs devos depends on along with devos for the new method to work because of an issue with the way `follows` works in flakes. This should be fixed by nixos/nix#4641 which would allow you to write a devos flake.nix in under 10 lines.

### DIFF
--- a/doc/start/use-mkdevos.md
+++ b/doc/start/use-mkdevos.md
@@ -1,0 +1,36 @@
+# Use mkDevos
+You can also use devos as a flake input and get the advantage of using
+nix flakes to sync with upstream changes in devos.
+
+To do this you will need the template and nix unstable. Edit your flake.nix to
+add devos as an input:
+```nix
+inputs = {
+  ...
+  devos.url = "github:divnix/devos";
+};
+```
+then replace the outputs section with this:
+```nix
+outputs = { self, devos, ... }: devos.lib.mkDevos { inherit self; };
+```
+
+Now you can update devos as you would any other flake input:
+```sh
+nix flake update --update-input devos
+```
+
+If you would like to change your directory's structure, you can pass
+arguments to `mkDevos` to outline where certain things are. You can take a 
+look at [evalDevosArgs](../../lib/devos/evalDevosArgs.nix) to learn about
+the arguments that it can take.
+
+> ##### Notes:
+> - All inputs used in devos must also be added to your flake, due to an
+>   [issue](https://github.com/NixOS/nix/pull/4641) with the `follows` attribute.
+> - Optionally, you can safely delete [lib](../../lib), [tests](../../tests) and
+>   the [core profile](../../profiles/core)
+> - You can append `/community` to the `devos.url` to get access to community modules
+>   which can be added in [extern](../../extern).
+> - To use community profiles, you will have to copy them to your tree
+> - If you use this method, you will no longer be able to edit devos internals.
--- a/doc/start/use-mkdevos.md
+++ b/doc/start/use-mkdevos.md
@@ -2,8 +2,13 @@
 You can also use devos as a flake input and get the advantage of using
 nix flakes to sync with upstream changes in devos.
 
-To do this you will need the template and nix unstable. Edit your flake.nix to
-add devos as an input:
+You can either use the default template or use the 'mkdevos' template which only
+includes the necessary files for `mkDevos` usage. It can be pulled with:
+```sh
+nix flake init -t github:divnix/devos#mkdevos
+```
+
+Once you have a template, edit your flake.nix to add devos as an input:
 ```nix
 inputs = {
   ...
@@ -15,10 +20,7 @@ then replace the outputs section with this:
 outputs = { self, devos, ... }: devos.lib.mkDevos { inherit self; };
 ```
 
-Now you can update devos as you would any other flake input:
-```sh
-nix flake update --update-input devos
-```
+Now you can sync with upstream devos changes just like any other flake input.
 
 If you would like to change your directory's structure, you can pass
 arguments to `mkDevos` to outline where certain things are. You can take a 
@@ -28,9 +30,8 @@ the arguments that it can take.
 > ##### Notes:
 > - All inputs used in devos must also be added to your flake, due to an
 >   [issue](https://github.com/NixOS/nix/pull/4641) with the `follows` attribute.
-> - Optionally, you can safely delete [lib](../../lib), [tests](../../tests) and
->   the [core profile](../../profiles/core)
 > - You can append `/community` to the `devos.url` to get access to community modules
->   which can be added in [extern](../../extern).
+>   and packages which can be added in [extern](../../extern).
+> - Many folders in devos will no longer be needed and can, optionally, be deleted.
 > - To use community profiles, you will have to copy them to your tree
 > - If you use this method, you will no longer be able to edit devos internals.
--- a/flake.nix
+++ b/flake.nix
@@ -28,68 +28,11 @@
       pkgs.inputs.nixpkgs.follows = "nixos";
     };
 
-  outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
-    let
-      inherit (self) lib;
-      inherit (lib) os;
-
-      extern = import ./extern { inherit inputs; };
-
-      multiPkgs = os.mkPkgs;
-
-      outputs = {
-        nixosConfigurations =
-          import ./hosts (nixos.lib.recursiveUpdate inputs {
-            inherit multiPkgs extern;
-            defaultSystem = "x86_64-linux";
-            lib = nixos.lib.extend (final: prev: {
-              dev = self.lib;
-            });
-          });
-
-        homeConfigurations = os.mkHomeConfigurations;
-
-        nixosModules =
-          let moduleList = import ./modules/module-list.nix;
-          in lib.pathsToImportedAttrs moduleList;
-
-        homeModules =
-          let moduleList = import ./users/modules/module-list.nix;
-          in lib.pathsToImportedAttrs moduleList;
-
-        overlay = import ./pkgs;
-        overlays = lib.pathsToImportedAttrs (lib.pathsIn ./overlays);
-
-        lib = import ./lib { inherit nixos self inputs; };
-
-        templates.flk.path = ./.;
-        templates.flk.description = "flk template";
-        defaultTemplate = self.templates.flk;
-
-        deploy.nodes = os.mkNodes deploy self.nixosConfigurations;
+    outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
+      let
+        lib = import ./lib { inherit self nixos inputs; };
+      in
+      lib.mkDevos {
+        inherit self;
       };
-
-      systemOutputs = utils.lib.eachDefaultSystem (system:
-        let pkgs = multiPkgs.${system}; in
-        {
-          checks =
-            let
-              tests = nixos.lib.optionalAttrs (system == "x86_64-linux")
-                (import ./tests { inherit self pkgs; });
-              deployHosts = nixos.lib.filterAttrs
-                (n: _: self.nixosConfigurations.${n}.config.nixpkgs.system == system) self.deploy.nodes;
-              deployChecks = deploy.lib.${system}.deployChecks { nodes = deployHosts; };
-            in
-            nixos.lib.recursiveUpdate tests deployChecks;
-
-          packages = utils.lib.flattenTreeSystem system
-            (os.mkPackages { inherit pkgs; });
-
-          devShell = import ./shell {
-            inherit self system;
-          };
-        }
-      );
-    in
-    nixos.lib.recursiveUpdate outputs systemOutputs;
 }
--- a/lib/attrs.nix
+++ b/lib/attrs.nix
@@ -21,7 +21,9 @@ rec {
       paths' = lib.filter (lib.hasSuffix ".nix") paths;
     in
     genAttrs' paths' (path: {
-      name = lib.removeSuffix ".nix" (baseNameOf path);
+      name = lib.removeSuffix ".nix"
+        # Safe as long this is just used as a name
+        (builtins.unsafeDiscardStringContext (baseNameOf path));
       value = import path;
     });
 
--- a/lib/devos/default.nix
+++ b/lib/devos/default.nix
@@ -11,6 +11,8 @@
 
   mkNodes = dev.callLibs ./mkNodes.nix;
 
+  mkSuites = dev.callLibs ./mkSuites.nix;
+
   mkProfileAttrs = dev.callLibs ./mkProfileAttrs.nix;
 
   mkPkgs = dev.callLibs ./mkPkgs.nix;
--- a/lib/devos/default.nix
+++ b/lib/devos/default.nix
@@ -7,7 +7,12 @@
       config = { allowUnfree = true; };
     };
 
+  importIf = path: if builtins.isPath path || builtins.isString path
+    then import path else path;
+
   profileMap = map (profile: profile.default);
+
+  evalDevosArgs = dev.callLibs ./evalDevosArgs.nix;
 
   mkNodes = dev.callLibs ./mkNodes.nix;
 
--- a/lib/devos/default.nix
+++ b/lib/devos/default.nix
@@ -11,6 +11,8 @@
 
   mkNodes = dev.callLibs ./mkNodes.nix;
 
+  mkHosts = dev.callLibs ./mkHosts.nix;
+
   mkSuites = dev.callLibs ./mkSuites.nix;
 
   mkProfileAttrs = dev.callLibs ./mkProfileAttrs.nix;
--- a/lib/devos/default.nix
+++ b/lib/devos/default.nix
@@ -7,8 +7,8 @@
       config = { allowUnfree = true; };
     };
 
-  importIf = path: if builtins.isPath path || builtins.isString path
-    then import path else path;
+  importIf = path: fallback:
+    if builtins.pathExists path then import path else fallback;
 
   profileMap = map (profile: profile.default);
 
--- a/lib/devos/default.nix
+++ b/lib/devos/default.nix
@@ -14,6 +14,8 @@
 
   evalDevosArgs = dev.callLibs ./evalDevosArgs.nix;
 
+  mkDevos = dev.callLibs ./mkDevos.nix;
+
   mkNodes = dev.callLibs ./mkNodes.nix;
 
   mkHosts = dev.callLibs ./mkHosts.nix;
--- a/lib/devos/evalDevosArgs.nix
+++ b/lib/devos/evalDevosArgs.nix
@@ -66,20 +66,23 @@ let
           apply = os.mkProfileAttrs;
           description = "path to user profiles folder";
         };
-        suites = mkOption {
-          type = inputAttrs;
-          default = importIf "${self}/suites"
-            ({...}: {user = {}; system = {};});
-          defaultText = "\${self}/suites";
-          apply = suites: os.mkSuites {
-            inherit suites;
-            inherit (config) profiles users userProfiles;
+        suites =
+          let
+            defaults = { user = {}; system = {}; };
+          in
+          mkOption {
+            type = inputAttrs;
+            default = importIf "${self}/suites" ({...}: defaults);
+            defaultText = "\${self}/suites";
+            apply = suites: defaults // os.mkSuites {
+              inherit suites;
+              inherit (config) profiles users userProfiles;
+            };
+            description = ''
+              function with inputs 'users' and 'profiles' that returns attribute 'system'
+              which defines suites passed to configurations as the suites specialArg
+            '';
           };
-          description = ''
-            function with inputs 'users' and 'profiles' that returns attribute 'system'
-            which defines suites passed to configurations as the suites specialArg
-          '';
-        };
         users = mkOption {
           type = path;
           default = "${self}/users";
@@ -136,17 +139,17 @@ let
             ## ${name}
             ${value.description}
 
-            ${optionalString (hasAttr "type" value ) ''
+            ${optionalString (value ? type) ''
               *_Type_*:
               ${value.type}
             ''}
-            ${optionalString (hasAttr "default" value) ''
+            ${optionalString (value ? default) ''
               *_Default_*
               ```
-              ${builtins.toJSON value.default or ""}
+              ${builtins.toJSON value.default}
               ```
             ''}
-            ${optionalString (hasAttr "example" value) ''
+            ${optionalString (value ? example) ''
               *_Example_*
               ```
               ${value.example}
--- a/lib/devos/evalDevosArgs.nix
+++ b/lib/devos/evalDevosArgs.nix
@@ -26,7 +26,10 @@ let
         };
         packages = mkOption {
           # functionTo changes arg names which breaks flake check
-          type = types.anything;
+          type = types.anything // {
+            check = builtins.isFunction;
+            description = "Nixpkgs overlay";
+          };
           default = importIf "${self}/pkgs" (final: prev: {});
           defaultText = "\${self}/pkgs";
           description = ''
--- a/lib/devos/evalDevosArgs.nix
+++ b/lib/devos/evalDevosArgs.nix
@@ -10,10 +10,8 @@ let
       cfg = config;
       inherit (cfg) self;
 
-      pathOr = x: with types; oneOf [ path x ];
       inputAttrs = types.functionTo types.attrs;
-
-          in
+    in
     {
       options = with types; {
         self = mkOption {
@@ -27,10 +25,9 @@ let
           description = "Path to directory containing host configurations";
         };
         packages = mkOption {
-          type = pathOr (functionTo inputAttrs);
-          default = "${self}/pkgs";
+          type = functionTo inputAttrs;
+          default = importIf "${self}/pkgs" (final: prev: {});
           defaultText = "\${self}/pkgs";
-          apply = importIf;
           description = ''
             Overlay for custom packages that will be included in treewide 'pkgs'.
             This should follow the standard nixpkgs overlay format - two argument function
@@ -38,38 +35,41 @@ let
           '';
         };
         modules = mkOption {
-          default = "${self}/modules/module-list.nix";
+          type = listOf anything;
+          default = importIf "${self}/modules/module-list.nix" [];
           defaultText = "\${self}/modules/module-list.nix";
-          type = pathOr (listOf anything);
-          apply = x: dev.pathsToImportedAttrs (importIf x);
+          apply = dev.pathsToImportedAttrs;
           description = "list of modules to include in confgurations";
         };
         userModules = mkOption {
-          default = "${self}/users/modules/module-list.nix";
+          type = listOf anything;
+          default = importIf "${self}/users/modules/module-list.nix" [];
           defaultText = "\${self}/users/modules/module-list.nix";
-          type = pathOr (listOf anything);
-          apply = x: dev.pathsToImportedAttrs (importIf x);
+          apply = dev.pathsToImportedAttrs;
           description = "list of modules to include in home-manager configurations";
         };
         profiles = mkOption {
           type = path;
           default = "${self}/profiles";
           defaultText = "\${self}/profiles";
+          apply = os.mkProfileAttrs;
           description = "path to profiles folder";
         };
         userProfiles = mkOption {
           type = path;
           default = "${self}/users/profiles";
           defaultText = "\${self}/users/profiles";
+          apply = os.mkProfileAttrs;
           description = "path to user profiles folder";
         };
         suites = mkOption {
-          type = pathOr inputAttrs;
-          default = "${self}/suites";
+          type = inputAttrs;
+          default = importIf "${self}/suites"
+            ({...}: {user = {}; system = {};});
           defaultText = "\${self}/suites";
-          apply = x: os.mkSuites {
+          apply = suites: os.mkSuites {
+            inherit suites;
             inherit (config) profiles users userProfiles;
-            suites = importIf x;
           };
           description = ''
             function with inputs 'users' and 'profiles' that returns attribute 'system'
@@ -80,19 +80,28 @@ let
           type = path;
           default = "${self}/users";
           defaultText = "\${self}/users";
+          apply = os.mkProfileAttrs;
           description = "path to folder containing user profiles";
         };
-        extern = mkOption {
-          type = pathOr inputAttrs;
-          default = "${self}/extern";
-          defaultText = "\${self}/extern";
-          apply = x: (importIf x) { inputs = inputs // self.inputs; };
-          description = ''
-            Function with argument 'inputs' with all devos and this flake's inputs.
-            The function should return an attribute set with modules, overlays, and
-            specialArgs to be included across devos
-          '';
-        };
+        extern =
+          let
+            defaults = {
+              modules = []; overlays = []; specialArgs = {};
+              userModules = []; userSpecialArgs = [];
+            };
+          in
+          mkOption {
+            type = inputAttrs;
+            default = importIf "${self}/extern" ({...}: defaults);
+            defaultText = "\${self}/extern";
+            # So unneeded extern attributes can safely be deleted
+            apply = x: defaults // (x { inputs = inputs // self.inputs; });
+            description = ''
+              Function with argument 'inputs' with all devos and this flake's inputs.
+              The function should return an attribute set with modules, overlays, and
+              specialArgs to be included across devos
+            '';
+          };
         overlays = mkOption {
           type = path;
           default = "${self}/overlays";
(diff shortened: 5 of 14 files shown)


lib/devos/mkDevos.nix
```diff
@@ -0,0 +1,78 @@
...
+, packages ? "${self}/pkgs"
+, modules ? import "${self}/modules/module-list.nix"
+, profiles ? "${self}/profiles"
+, suites ? import "${self}/suites"
+, users ? "${self}/users"
+, extern ? import "${self}/extern"
+, overlays ? "${self}/overlays"
+, overrides ? import "${self}/overrides"
+, ...
+}:
```
Pacman99, line 17: These arguments could be typechecked and documented with the module system.

lib/devos/mkDevos.nix
```diff
@@ -0,0 +1,78 @@
+{ self, nixos, inputs, ... }:
+let
+  devos = self;
```
Pacman99, line 3: This line is a bit wierd, but essentially the first self is from callLib which is always called from devos itself. The self below is what allows mkDevos to be used on other flakes.

lib/devos/mkSuites.nix
```diff
@@ -0,0 +1,25 @@
...
+let
+  inherit (dev) os;
+
+  args = {
+    profiles = os.mkProfileAttrs (toString profiles);
+    users = os.mkProfileAttrs (toString users);
+  };
+
+  definedSuites = (suites { inherit (args) profiles users; }).system;
+
```
Pacman99, line 13: Had to format some things a bit odd to prevent infinite recursion

nrdxp:
Very interesting. I had another exhausting day, so I won't even try to review this right now. But I'll definitely have a good look at this soon.

blaggacao (commented):
This totally satisfies my desire for _clear interfaces_.

I've taken a first read and I find it really easy to reason about it.

It is really comfortable and totally magic-less to have `mkDevos.nix` give an overview of the default file system layout in only aboy 10 lines from the function arguments.

I have a few associations, but I'll keep them for opening new issues and discuss them individially / incrementally once this general direction is merged.

gytis-ivaskevicius:
Hey, I am the guy from Reddit - I have to say, this really does look like my project :D Good stuff tho, hopefully, this will get merged! :)

blaggacao:
@Pacman99 totally open ended question: what benefits and drawbacks would you see in making this a `flake-lib-plus-plus` repo within `divnix` orga? Then we can &mdash; for the fun and [crosspolination](https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/2) &mdash; rebase on [`flake-utils-plus`](https://github.com/gytis-ivaskevicius/flake-utils-plus).

Pacman99:
>@Pacman99 totally open ended question: what benefits and drawbacks would you see in making this a flake-utils-plus-plus repo within divnix orga?

Actually now that you mention that, I was actually considering setting up a fork of devos that exported all its features as a lib function. I would really like to not do that, but if I did I would keep it as close to devos as possible and just add a lib function layer. But it would be much better if it was in devos instead. I mainly wasn't sure if you all wanted devos to go in this direction of being a lib function rather than a template. This PR attempts to make a compromise by doing both. Move everything to lib but maintain folder structure/template. 
If this PR was done as a fork, it would be really hard to maintain. And I think moving the logic to lib is a change that would really improve devos, even if we decide not to include the `mkDevos` part. 

Also there is one feature I like from `flake-utils-plus` is you can pass builders and change how things are exported. It would be cool if you can pass some sort of overlay to `mkDevos`. Perhaps you could override lib functions and change how things are implemented.

blaggacao:
> I mainly wasn't sure if you all wanted devos to go in this direction of being a lib function rather than a template. This PR attempts to make a compromise by doing both.

Very wise choice. I agree, before v1.0, or even beyond that point, it is a feature to be able to hack one's own `mkNetBoot = netbootConfigurationBuilder` by means of commit trees rather than nix overlays (upstream PR is litterally just one click away).

nrdxp:
it's all you @Pacman99
delegate+

lib/devos/mkDevos.nix
```diff
@@ -0,0 +1,78 @@
...
+      extern = external;
+      suites = allSuites;
+      dir = hosts;
+    };
+
+    nixosModules = lib.pathsToImportedAttrs modules;
+
+    overlay = import packages;
+    #overlays = lib.pathsToImportedAttrs (lib.pathsIn (builtins.toPath overlays));
+    overlays = lib.pathsToImportedAttrs (lib.pathsIn ../../overlays);
```
Pacman99, line 42: I seem to get a problem with this part. The commented line would be what we actually want as to avoid directly referencing paths in mkDevos and lib functions. But that seems to cause an error that the string 'nixFlakes' cannot refer to a store path. I don't really know much about string contexts(as I believe their called), so I don't know whats wrong. I can't figure out why a string isn't allowed to contain store paths within a flake.

nrdxp: iirc, I ran into this same error. Perhaps we should report this upstream, as it seems like sometimes this works and sometimes it doesn't, and the reason why isn't exactly clear.

Pacman99: I'm glad to see its not just me. I'll see if I can file a report. I want to try and reproduce it with a simpler flake, so I can identify what exactly the issue is.

Pacman99: The nixos update seems to have resolved this issue. The overlays line works for me now.

lib/devos/mkProfileAttrs.nix
```diff
@@ -27,7 +27,7 @@ let mkProfileAttrs =
     f = n: _:
       lib.optionalAttrs
         (lib.pathExists "${dir}/${n}/default.nix")
-        { default = /. + "${dir}/${n}"; }
+        { default = builtins.toPath "${dir}/${n}"; }
```
Pacman99, line 30: This is a deprecated function, but without this I get an error that store paths in strings can't be appended to paths.
I tried `builtins.path` too and that didn't work. I think this could be reported upstream, I'm just not sure what to report.

lib/devos/mkDevos.nix
```diff
@@ -0,0 +1,89 @@
...
+      path = builtins.toPath self;
+      description = "flk template";
+    };
+      /*mkdevos =
+        # Folders that aren't necessary with mkDevos method
+        let excludes = [ "lib" "tests" "nix" ]; in
+        {
+          path = builtins.filterSource (f: ! (builtins.elem f excludes)) self;
+          description = "template used for mkdevos";
+        }; */
```
Pacman99, line 61: This is an attempt at making a template for the new method of using devos. It doesn't work due to filterSource not accepting store paths. Its not really necessary, the alternative is just documentation on how to use the mkDevos function and folders that are not necessary.

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
This is ready for final review. Added a doc for using mkdevos and cleaned up all the code. Keep in mind, if merged, this will likely create merge conflicts in most existing PR's.

lib/devos/mkProfileAttrs.nix
nrdxp: does `toString` work?

Pacman99: Not sure what you mean. It is already a string. But I did try `toString /. + ...`. That didn't work, but it is redundant since it defeats the point of the original change.

lib/devos/mkDevos.nix
nrdxp: did you decide against this or saving for another PR?

lib/devos/mkDevos.nix
Pacman99: Just realized that this is only no longer a problem because the overlay was removed. Not really sure what to do about the the overlay problem.

lib/devos/mkDevos.nix
Pacman99: Actually I'm working on it right now and I might just include it in this PR. I'm not sure if thats the best idea though.

Pacman99:
Nevermind, this might need some more review. But with nixpkgs master features available I went ahead and setup argument evaluation using the module system. I'm sorry if this makes this PR harder to review, as it has grown pretty large, but it makes everything much cleaner. 

Using module system apply keys I was able to directly store all of devos data within the evaluated modules. So `mkDevos` has to do less work, and its easy to see how each argument is evaluated.

nrdxp:
Once this _is_ merged, we should probably tag 0.9.0

Pacman99:
Okay various updates, I added a `genDoc` function to `evalDevosArgs` that creates markdown documentation. This is custom made, so I the formatting could use some review. You can now build option documentation with 
`nix build .#devosOptionsDoc`. I'm not sure where to put it in tree, I'm thinking of putting it in `lib`.

Pacman99:
Dropped the `pathOr` type so all options are explicitly type checked. And instead the importing of paths is done as part of the default with the new `importIf` function, which allows for fallbacks. This has the added benefit of many options not being required, for example, users can safely delete the modules folder and due to the fallback in `importIf` no error will be triggered. 
I also moved more of the implementation logic for each option to their relevant apply keys: profiles, users, and userProfiles get directly evaluated through the apply key. So mkSuites does less work.

Pacman99:
Update: created a mkdevos template with a list of excluded folders that aren't needed when using mkdevos. And some general doc cleanups.

Pacman99:
The only thing that I have left for this PR is to fix the overlays, so it doesn't break community.
Outside of that, I think this is fully for ready for review - I don't plan on changing anything else. @gytis-ivaskevicius If you could take a look too that would be great - I can't request your review on github ui. 
Also the function might be renamed to mkDevosFlake(@blaggacao idea), but I'm not yet sure about that. And possibly a test might be added for `mkDevos`.

lib/devos/mkDevos.nix
Pacman99: this was an issue with string contexts, so I updated pathsToImportedAttrs to discard string context in the names

blaggacao:
> Also the function might be renamed to mkDevosFlake(@blaggacao idea), but I'm not yet sure about that.

`numtide/devshell` went from `devshell.mkDevshell`: "devshell, make a devshell" to `devshell.mkShell`: "devshell, make a shell of yours".

Reason: the authors probably believed, as I do, that there _are_ people who notice the fine matices of overzealous "branding" and percieve it as obnauxious.

Code becomes more concise, if one is able to read it out in plain english, hence:

`devos.mkDevos`: "devos, make a dev OS", in addition to potential overzealous beanding due to repetition, is semantically not correct. This function does not make a single operating system of type "dev". Correct is: "devos, make an environment of yours" or a more scoped "devos, make a flake of yours", but if we want to impose ourselves a little more "devos, make a _Devos_ environment" or "devos, make a _Devos_ flake".

I'd vote for "devos, make a flake if yours" which has most informational value per char and is in line with stances of `numtide/devshell` & `flake-utils-plus`.

Note that, with flakes as opposed to traditional global `pkgs` namespace, the caller will most likely have the `devos` namespace in scope, hence duplicating the repo brand is a statement, we might not actually want to make.

blaggacao (changes requested):
Ha, my review from two days ago survied battery die-out in the cache...

I'll request some changes in naming and docs to polish this off.

The docs might need a few iterations, still. I'm concerned with just throwing "random" options without super clear use cases or context at users in plain _nix-ecosystem-style_. To reach our goals, this is one of the in my opinion bad practices in the nix ecosystem, I don't think we should inherit.

The docs should answer more questions, than they generate. Here, I'm especially concerned to have a cryptic "Use mkDevos" in the index in absence of a clear self-identifying use-case in the title, like "How to get started" or "Use devos as a library!" &mdash; with exclamation mark it is a _recommendation_ (with impetus) and I think in line what we actually want to recommend (with impetus) in the future.

Then we could elaborate around that central theme "use it as a library" as a three paragrapher with a structure around the lines of either of e.g.
- problem &mdash; solution &mdash; benefit
- what? &mdash; now what? &mdash; so what?
- benefit &mdash; solution &mdash; example

I could offer to have a take on the docs, if you want, but it would be some days into the future still, until I have again full dominion over my _working_ workstation. Let me know.

lib/devos/mkSuites.nix
```diff
@@ -0,0 +1,24 @@
+{ lib, dev, ... }:
```
blaggacao, line 1: As promised above, I wanted to suggest to define suites within flake.nix or _at least_ within top level suites.nix.

Suites are a higher level aggregate of profiles and users, their definition is ultrashort. I don't see a point &mdash; by reason of analogy &mdash; to burry them in a folder (at the same level as profiles).

blaggacao (changes requested):
Follow up with went was missing above...

lib/devos/evalDevosArgs.nix
```diff
@@ -0,0 +1,103 @@
+{ self, dev, nixos, inputs, ... }:
+
+{ args }:
+let
+  argOpts = with nixos.lib; { config, ... }:
+    let
+      inherit (dev) os;
+      inherit (os) importIf;
```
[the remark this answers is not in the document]
blaggacao: I'm also not sold on not declaring it locally as long as it is not of actual use in another place.

lib/devos/evalDevosArgs.nix
```diff
@@ -0,0 +1,166 @@
...
+      inputAttrs = types.functionTo types.attrs;
+    in
+    {
+      options = with types; {
+        self = mkOption {
+          type = addCheck attrs nixos.lib.isStorePath;
+          description = "The flake to create the devos outputs for";
+        };
+        hosts = mkOption {
+          type = path;
```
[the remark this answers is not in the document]
blaggacao: An alternative is to layer another funktion:

`devos.mkDevosFlake` over `devos.mkFlake`, but we really forego a stripped down flake.nix "index"-service if we don't _always_ specify paths directly within the flake.nix.

So we shoudn't do a `mkdDevosFlake` / `mkDevos` at all: the "devos" part in `mkDevos` would actually become a template. That is much more inclusive of potential future library users.

blaggacao:
_I'm sorry for my convoluted feedback. If it results impossible / inefficient to consume, please split up this PR into its atomic parts, and I'll re-apply a more focused feedback on each increment._

After reverberation in my thoughts, I came to the conclusion, we ought to make any devos library agnostic of any file hierarchy layout. For indexing purpose, file hierachy needs to be declared in (a henceforth stripped down) top level `flake.nix`

File layout is best encoded as "batteries included but removable" within a template. An abstract devos's library is also attractive to more people (wink at @gytis-ivaskevicius 's impetus) and thereby promotes "consensus".

`devos`, then, furthermore provides library functions to make it easy to work with our heredary (or similar) file layout.

A rough api design for library users would look like this:

```
devos.mkFlake
devos.mkHost
...
devos.lib.mkAttributesFromPaths
...
```

(No point in putting `mkFlake` under `devos.lib` since external consumers don't use _devos-as-library_ as anything else than a _library_.)

We then could reflect that in devos-the-template and adopt it for in-tree devos-the-blueprint (aka community): `inputs.devos.path = ./lib`.

lib/devos/default.nix
```diff
@@ -7,10 +7,21 @@
       config = { allowUnfree = true; };
     };
 
+  importIf = path: if builtins.isPath path || builtins.isString path
```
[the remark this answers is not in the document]
blaggacao: Just realized, its different here. But as elaborated in conjunction with my other comments, it probably should not be.

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,36 @@
...
+add devos as an input:
+```nix
+inputs = {
+  ...
+  devos.url = "github:divnix/devos";
+};
+```
+then replace the outputs section with this:
+```nix
+outputs = { self, devos, ... }: devos.lib.mkDevos { inherit self; };
```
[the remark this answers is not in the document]
blaggacao: `devos.mkFlake` for external devos-as-a-library users.

lib/devos/evalDevosArgs.nix
```diff
@@ -0,0 +1,103 @@
...
+            function with inputs 'users' and 'profiles' that returns attribute 'system'
+            which defines suites passed to configurations as the suites specialArg
+          '';
+        };
+        users = mkOption {
+          type = path;
+          default = "${self}/users";
+          description = "path to folder containing user profiles";
+        };
+        extern = mkOption {
```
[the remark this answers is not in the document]
blaggacao: Yeah, this is part of the prime `devos` api,  we should refrain from putting it into a badly named `extern` config object.
Why 2 levels of api?!?

lib/devos/default.nix
```diff
@@ -7,10 +7,21 @@
       config = { allowUnfree = true; };
     };
 
+  importIf = path: fallback:
```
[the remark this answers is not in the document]
blaggacao: Big reverbaration step of mine, see other comment about my thought process, here.

blaggacao:
Note:

With devos-a-library there is still a need for `./lib` in the template and wiring as in "my environment's lib".

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,37 @@
...
+Once you have a template, edit your flake.nix to add devos as an input:
+```nix
+inputs = {
+  ...
+  devos.url = "github:divnix/devos";
+};
+```
+then replace the outputs section with this:
+```nix
+outputs = { self, devos, ... }: devos.lib.mkDevos { inherit self; };
```
gytis-ivaskevicius, line 20: Consider instead of specifying individual changes be like "this is how your flake should look like" and users can add extra attributes if necessary

flake.nix
```diff
@@ -28,68 +28,11 @@
...
-        lib = import ./lib { inherit nixos self inputs; };
-
-        templates.flk.path = ./.;
-        templates.flk.description = "flk template";
-        defaultTemplate = self.templates.flk;
-
-        deploy.nodes = os.mkNodes deploy self.nixosConfigurations;
+    outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
+      let
+        lib = import ./lib { inherit self nixos inputs; };
```
gytis-ivaskevicius, line 33: Just a little idea for the future - separate the library implementation from this template repository

(probably not relevant as part of this PR)

lib/attrs.nix
```diff
@@ -21,7 +21,9 @@ rec {
       paths' = lib.filter (lib.hasSuffix ".nix") paths;
     in
     genAttrs' paths' (path: {
```
gytis-ivaskevicius, line 23: In Nix usually, collection itself (in this case `paths'`) is specified as the last argument. While we are at it - maybe it's time to change it as well?

lib/devos/evalDevosArgs.nix
```diff
@@ -0,0 +1,166 @@
...
+          defaultText = "\${self}/hosts";
+          description = "Path to directory containing host configurations";
+        };
+        packages = mkOption {
+          # functionTo changes arg names which breaks flake check
+          type = types.anything // {
+            check = builtins.isFunction;
+            description = "Nixpkgs overlay";
+          };
+          default = importIf "${self}/pkgs" (final: prev: {});
```
gytis-ivaskevicius, line 33: It makes me wonder if users won't get confused at first that something gets picked up by just creating a folder (without manually declaring it)

lib/devos/default.nix
```diff
@@ -7,10 +7,21 @@
...
+    if builtins.pathExists path then import path else fallback;
```
gytis-ivaskevicius, line 11: Won't it crash if `path` does not contain `default.nix` file?

lib/devos/mkDevos.nix
```diff
@@ -0,0 +1,72 @@
...
+    deploy.nodes = os.mkNodes deploy self.nixosConfigurations;
+  };
+
+  systemOutputs = utils.lib.eachDefaultSystem (system:
+    let pkgs = multiPkgs.${system}; in
+    {
+      checks =
+        let
+          tests = nixos.lib.optionalAttrs (system == "x86_64-linux")
+            (import "${devos}/tests" { inherit self pkgs; });
```
gytis-ivaskevicius, line 55: other platforms not supported????

lib/devos/mkHosts.nix
```diff
@@ -48,15 +38,15 @@ let
...
       ];
 
       nixpkgs.pkgs = lib.mkDefault multiPkgs.${config.nixpkgs.system};
 
       nix.registry = {
         devos.flake = self;
         nixos.flake = nixos;
-        override.flake = override;
+        override.flake = inputs.override;
       };
```
gytis-ivaskevicius, line 50: Consider something like this https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/staging/modules/saneFlakeDefaults.nix

lib/devos/mkHosts.nix
```diff
@@ -48,15 +38,15 @@ let
       nix.nixPath = [
         "nixpkgs=${nixos}"
         "nixos-config=${self}/compat/nixos"
-        "home-manager=${home}"
+        "home-manager=${inputs.home}"
       ];
```
gytis-ivaskevicius, line 42: I am getting rid of NIX_PATH in an upcoming release of `flake-utils-plus`, it does not seem to be particularly relevant in the world of Nix 3.0

Feel free to keep it - tho I'd suggest checking this one out https://discourse.nixos.org/t/i-feel-like-nix-path-creates-more-issues-than-it-fixes/12110/9

lib/devos/mkSuites.nix
```diff
@@ -0,0 +1,24 @@
...
+    inherit (args) users profiles userProfiles;
+  };
+
+  allProfiles =
+    let defaults = lib.collect (x: x ? default) profiles;
+    in map (x: x.default) defaults;
+
+  allUsers =
+    let defaults = lib.collect (x: x ? default) users;
+    in map (x: x.default) defaults;
```
gytis-ivaskevicius, line 17: Consider defining `myFunc = it: lib.collect (x: x ? default) it;`

shell/default.nix
```diff
@@ -1,9 +1,8 @@
 { self ? (import ../compat).defaultNix
 , system ? builtins.currentSystem
+, pkgs ? import <nixpkgs> {}
```
gytis-ivaskevicius, line 3: ???? is this necessary? `nix develop` should take care of nixpkgs, not NIX_PATH

gytis-ivaskevicius (approved):
Can't wait to see this merged :)

lib/devos/evalDevosArgs.nix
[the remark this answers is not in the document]
Pacman99: I could just declare it in evalDevosArgs, its really only useful there.
Still thinking of names, `importWithFallback` is along the lines of what I'll probably go with.

Pacman99:
Thanks @gytis-ivaskevicius and @blaggacao for all the review! As always the feedback is amazing. Following @blaggacao suggestion, I'm splitting up all these changes into separate PR's, which can each be merged by themselves. So I'll probably end up marking this one as a draft and make a new PR for just the change that involves moving flake logic into lib. 
I will still look at all the review here, and try to address it in the other PR's. Also some of @gytis-ivaskevicius's suggestions were more of a critique on parts of devos itself, which is great and definitely something we should address but it will probably be better done in separate PR's. They did make some great points, and we should get around to changing them soon.

shell/default.nix
Pacman99: Adding the pkgs part is necessary so that the shell implementation doesn't try to create its own instance of pkgs and so it can be called within lib.
The `import <nixpkgs> {}` part is there because the shell folder is also there for backwards compatibility. So you can still do `nix-shell`. Without it, nix-shell would error out.

lib/devos/mkSuites.nix
Pacman99: Yeah good idea, since I'm already changing the suites implementation logic in #216 I can also address this.

lib/devos/mkHosts.nix
Pacman99: I have to agree, NIX_PATH is not the best. I don't know if we will remove this anytime soon though, one of devos goals is to help with the transition to flakes. This part should definitely be moved eventually.

lib/devos/mkHosts.nix
Pacman99: Yeah I saw that earlier when looking through flake-utils-plus. I've actually had an issue with that though, and I've been meaning to bring it up in devos at some point. When you evaluate your flake after building the registry, the lock file will change to reference paths in your nix store instead of the url's to the inputs. And this sort of reduces `reproducability` since you can't evaluate your flake on other machines as easily. I ran across this while setting up CI for my server, it kept failing since nix was trying to reference store paths that didn't exist in the CI machine.

lib/devos/mkDevos.nix
Pacman99: Unfortunately not yet, the tests in devos need a bit of reworking for that to happen. But other platforms will be checked using `deploy-rs` checks, so its not terrible.

lib/devos/default.nix
Pacman99: Yes, but this is supposed to be a thin wrapper around nix's `import` and that would crash if there is no default.nix.

lib/devos/evalDevosArgs.nix
Pacman99: Good point, I should be able to address this soon when switching to `mkFlake` which will be more general and then devos will use `mkFlake` on itself. That way the `importIf` is not necessary either.. @blaggacao I think this is a good example of the corruption you mentioned.

flake.nix
Pacman99: I think I would be for that. @nrdxp @blaggacao thoughts?

doc/start/use-mkdevos.md
Pacman99: I like that idea. I think docs should be its own PR too, discussion for that can happen on its own.

lib/devos/mkSuites.nix
Pacman99: Well this function doesn't actually declare suites, it just processes them. suites are still declared in top level and can be declared within the flake. The goal here was to prevent suites from having to do the work of importing profiles and processing them.

lib/devos/evalDevosArgs.nix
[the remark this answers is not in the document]
Pacman99: Will address in planned `mkFake` PR. So only devos encodes its own filesystem api within its own flake.nix, but its not encoded within lib.

lib/devos/default.nix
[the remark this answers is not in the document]
Pacman99: Yeah with the planned `mkFlake` I don't think the 'importIf' function will be necessary. I really only made it to maintain backwards compatibility so mkDevos could be used on devos itself. But I see now why that would be better done outside of lib.

lib/attrs.nix
```diff
@@ -21,7 +21,9 @@ rec {
...
-      name = lib.removeSuffix ".nix" (baseNameOf path);
+      name = lib.removeSuffix ".nix"
+        # Safe as long this is just used as a name
+        (builtins.unsafeDiscardStringContext (baseNameOf path));
```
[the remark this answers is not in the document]
Pacman99: see pr that changes `pathsToImportedAttrs` I put reasoning there.

lib/devos/evalDevosArgs.nix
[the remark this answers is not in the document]
Pacman99: I see your point, but I would also prefer to address things like this in a separate discussion/PR. For now I want mkDevos/mkFlake to just be a mirror of the current devos api. So that way we can separate discussion and lib implementation.

lib/devos/evalDevosArgs.nix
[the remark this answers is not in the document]
Pacman99: Should not be necessary with `mkFlake`

lib/devos/evalDevosArgs.nix
```diff
@@ -0,0 +1,103 @@
+{ self, dev, nixos, inputs, ... }:
```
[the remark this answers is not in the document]
Pacman99: I don't like dev either, it is kind of confusing. I'm not sure how to fix it though outside of better documentation. Its just a side effect of needing to export lib as `lib` in the flake, but also requiring a way to access devos lib within other devos lib functions.

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,36 @@
...
+
+> ##### Notes:
+> - All inputs used in devos must also be added to your flake, due to an
+>   [issue](https://github.com/NixOS/nix/pull/4641) with the `follows` attribute.
+> - Optionally, you can safely delete [lib](../../lib), [tests](../../tests) and
+>   the [core profile](../../profiles/core)
+> - You can append `/community` to the `devos.url` to get access to community modules
+>   which can be added in [extern](../../extern).
+> - To use community profiles, you will have to copy them to your tree
+> - If you use this method, you will no longer be able to edit devos internals.
```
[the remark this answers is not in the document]
Pacman99: I would appreciate being told this, just to make it clear the cons of using mkDevos.

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,36 @@
...
+the arguments that it can take.
+
+> ##### Notes:
+> - All inputs used in devos must also be added to your flake, due to an
+>   [issue](https://github.com/NixOS/nix/pull/4641) with the `follows` attribute.
+> - Optionally, you can safely delete [lib](../../lib), [tests](../../tests) and
+>   the [core profile](../../profiles/core)
+> - You can append `/community` to the `devos.url` to get access to community modules
+>   which can be added in [extern](../../extern).
+> - To use community profiles, you will have to copy them to your tree
```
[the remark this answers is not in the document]
Pacman99: Well its because profiles are not exported anywhere in the flake. I don't see the point in adding the `why` to docs. And also at some point we could try to export profiles under `nixosModules` to fix this.

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,36 @@
...
+
+If you would like to change your directory's structure, you can pass
+arguments to `mkDevos` to outline where certain things are. You can take a 
+look at [evalDevosArgs](../../lib/devos/evalDevosArgs.nix) to learn about
+the arguments that it can take.
+
+> ##### Notes:
+> - All inputs used in devos must also be added to your flake, due to an
+>   [issue](https://github.com/NixOS/nix/pull/4641) with the `follows` attribute.
+> - Optionally, you can safely delete [lib](../../lib), [tests](../../tests) and
```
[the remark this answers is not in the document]
Pacman99: I did provide a different template, its called `mkdevos`.

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,36 @@
...
+```sh
+nix flake update --update-input devos
+```
+
+If you would like to change your directory's structure, you can pass
+arguments to `mkDevos` to outline where certain things are. You can take a 
+look at [evalDevosArgs](../../lib/devos/evalDevosArgs.nix) to learn about
+the arguments that it can take.
+
+> ##### Notes:
```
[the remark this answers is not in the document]
Pacman99: My thoughts too, the docs need some work

doc/start/use-mkdevos.md
```diff
@@ -0,0 +1,36 @@
+# Use mkDevos
+You can also use devos as a flake input and get the advantage of using
```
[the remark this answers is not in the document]
Pacman99: I think for now it should be documented as a 'next step' and another option. My reasoning is that there is some work to switch to the `mkDevos` method, and it's not an easy change.

lib/devos/evalDevosArgs.nix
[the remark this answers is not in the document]
nrdxp: perhaps simply renaming to `osLib` or `devLib`. I really just went with `dev` to keep it as convenient as `lib`, i.e. three chars.

flake.nix
nrdxp: I'm definitely up for it.

lib/devos/evalDevosArgs.nix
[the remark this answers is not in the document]
Pacman99: I personally think dev is ok, my main issue is the fact that lib in some places refers to devos lib and in other places refers to nixos lib. perhaps we can just standardize dev and use it everywhere even in flake.nix. So dev always refers to devos lib and lib always refers to nixos lib.

flake.nix
blaggacao: Definitly, in the coming days, until 1.0 we might need the benefit of atomic spanning PRs, still, though.

shell/default.nix
blaggacao: Maybe we can use the back-compat flake (anti corruption layer) which iirc handles also shell back-compat for this class of use cases.

Pacman99: this has been addressed in #217 the solution has some potential for corruption, but only when using old `nix-shell`, so I don't think its all that bad.

Pacman99:
closing in favor of #218

PR #211, opened 2021-03-25
home-manager fixes

nrdxp:
- [x] integrate nix-community/home-manager#1880 (to avoid duplicate commits, we'll wait a bit to see if this gets merged)
- [x] don't use all of `systemPackages` for `flk home` 
- [x] test home-manager `activationPackage` and `flk home` as part of CI

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -118,11 +118,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1616318638,
-        "narHash": "sha256-E6ABXtzw6bHmrIirB1sJL6S2MEa3sfcvRLzRa92frCo=",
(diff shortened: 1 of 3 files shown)


Pacman99:
Also considering that this issue wasn't caught originally, perhaps add the activationPackage for nixos user to checks.

Pacman99:
If none of the hm PR's get merged soon you could also use the `extraSpecialArgs` option to fix the issue. That way no change in home-manager upstream is required. Since devos always sets `useGlobalPackages` that would be an okay fix.

nrdxp:
Ready for preliminary review.

Pacman99 (approved):

PR #212, opened 2021-03-25
lib/pathsIn: don't error if path doesn't exist

Pacman99:
Really simple, just don't error if path passed to pathsIn doesn't exist. This doesn't affect much, since the majority of folders in devos end up getting referenced in some way or another outside of a pathsIn call. But it will  help for users of mkDevos, so they can safely delete folders they don't want to use. 
Can test with current overlays folder in deovs, that only gets imported through a pathsIn call.

### DIFF
--- a/lib/lists.nix
+++ b/lib/lists.nix
@@ -4,5 +4,6 @@
     let
       fullPath = name: "${toString dir}/${name}";
     in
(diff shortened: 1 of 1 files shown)


nrdxp (approved):
LGTM

PR #215, opened 2021-03-26
lib/pathsToImportedAttrs: discard string context in the names to prevent flake check error

Pacman99:
So I learned about this myself while working on it. Nix has a thing called string context, where strings refer to various store paths that they used to contain. This is normally a good thing to guarantee purity, but flake outputs cannot contain store paths. So when `pathsToImportedAttrs` is passed a list of store paths, and is then used for a flake output, we get a very confusing error that the string refers to a store path - even though it doesn't look like it does! 
I think its good to review this as a separate Pr, since it requires calling an unsafe nix builtin.

### DIFF
--- a/lib/attrs.nix
+++ b/lib/attrs.nix
@@ -21,7 +21,9 @@ rec {
       paths' = lib.filter (lib.hasSuffix ".nix") paths;
     in
     genAttrs' paths' (path: {
-      name = lib.removeSuffix ".nix" (baseNameOf path);
+      name = lib.removeSuffix ".nix"
+        # Safe as long this is just used as a name
+        (builtins.unsafeDiscardStringContext (baseNameOf path));
       value = import path;
     });
 


lib/attrs.nix
```diff
@@ -21,7 +21,9 @@ rec {
       paths' = lib.filter (lib.hasSuffix ".nix") paths;
     in
     genAttrs' paths' (path: {
-      name = lib.removeSuffix ".nix" (baseNameOf path);
+      name = lib.removeSuffix ".nix"
+        # Safe as long this is just used as a name
+        (builtins.unsafeDiscardStringContext (baseNameOf path));
```
Pacman99, line 26: Here is where even though we chop of the nix store part of the string with `baseNameOf`, the string actually still contains the store path. So we need to discard string context. This should be safe as long as the name is always used as ... a name.

nrdxp (approved):
bors r+

Issue #136, opened 2021-02-23
Profiles imports are brittle, causing failure if imported twice


ymarkus:
## Expected Behavior
If I set `users.users.<name>.shell` in the `users/<name>/default.nix` file, there should be no conflicts anywhere (happens for any user).

## Current Behavior
Building the system configuration errors out with following:
for `users.users.<name>.shell = pkgs.zsh`:
```
The unique option `users.users.<name>.shell' is defined multiple times, in:
 - <unknown-file>
 - <unknown-file>.
```

## Possible Solution

Maybe I have to set it somewhere else? Can't find anything in the documentation...

nrdxp:
I can't reproduce this error in either core or community from a fresh checkout. What revision are you on? The only place I set the default shell for users is: https://github.com/divnix/devos/blob/community/profiles/develop/zsh/default.nix#L9

This doesn't conflict with users.users.<name>.shell though, as I tested it explicitly, and this is only set if your using the community branch and develop or zsh profile. You might want to grep around and see if you accidentally set it twice somewhere, or perhaps this was an issue with an older revision but is now resolved. Worst case scenario, you can always force the setting with `mkForce`.

I'll leave this open for a bit to help you find a solution.

ymarkus:
OK, I've found the error. It seems like if you import a profile twice, it does weird things. Here's what happened:

1. `users.<name>` get imported in core:
```nix
core = [ users.nixos users.root  ];
```
2. `core` gets imported in `graphics`:
```nix
graphics = core ++ [ graphical ];
```
3. `graphics` get imported in `pc` & `gaming`:
```nix
pc = graphics ++ [ develop networking.samba printing systems.pc virt zfs ];
gaming = graphics ++ [ graphical.games ];
```
4. `pc` & `gaming` get imported in `NixOS.nix`:
```nix
imports = [...] ++ suites.pc ++ suites.gaming;
```
So everything in the profiles `core` & `graphical` got imported twice. That's why mkForce also didn't work.

Maybe a hint in the manual would be fitting if this is the case?

nrdxp:
Good catch, we don't want this biting us in the future. Perhaps the profiles attribute set in suites should just contain paths to the profile, instead of importing it outright. The default import logic prevents this sort of error when importing the same path twice, so we might want to just leverage that.

ymarkus:
Yeah, I just found out it can also happen when importing a profile inside suites and then importing it inside another profile like this:
```nix
{
  suites = with profiles; rec {
[...]

PR #216, opened 2021-03-26
Move host and suite implementation logic to lib

Pacman99:
This is just the `mkSuites` and `mkHosts` part of the `mkDevos` PR. I would like to avoid changing mkSuites and mkHosts api, to make it easier to rebase changes into the mkDevos branch. But if necessary we can change them. And to that end, there is some more logic added to the flake.nix now which would ideally not be there if devos was meant to be a template. But since the goal is to move towards a lib function with template, this is just a step in that direction.


flake.nix
```diff
@@ -35,17 +35,21 @@
...
-          import ./hosts (nixos.lib.recursiveUpdate inputs {
-            inherit multiPkgs extern;
-            defaultSystem = "x86_64-linux";
-            lib = nixos.lib.extend (final: prev: {
-              dev = self.lib;
-            });
-          });
+        nixosConfigurations = os.mkHosts {
+          dir = "${self}/hosts";
+          overrides = import ./overrides;
```
Pacman99, line 50: All paths here have to passed as strings or nix says paths aren't allowed to refer to the flake store path. Makes no sense to me since those paths are within the flake, but this is the workaround.

lib/devos/mkProfileAttrs.nix
```diff
@@ -27,7 +27,7 @@ let mkProfileAttrs =
     f = n: _:
       lib.optionalAttrs
         (lib.pathExists "${dir}/${n}/default.nix")
-        { default = /. + "${dir}/${n}"; }
+        { default = builtins.toPath "${dir}/${n}"; }
```
Pacman99, line 30: store paths cannot be appended to paths. This is a deprecated function(I think?), but it is the only way I found of converting a store path to a 'path' type. And we need these as paths for the `disabledModules` part of `isoConfig` to work.

nrdxp: `disabledModules` should also work by using a partial path string relative to the systems nixpkgs/nixos/modules directory. So maybe this isn't needed after all.

Pacman99: But we are trying to disable paths in devos, so we can't pass anything relative to nixpkgs/nixos/modules. The other option is to set modulesPath to "profiles/" but I'm not sure if that breaks other things in the module system. I think its better to just include this change for now, to get these functions working. And in the future work out a way to pass these paths as strings and still get it to work with `disabledModules`. I want to avoid changing isoConfig in this PR.

flake.nix
blaggacao: That means even `./host` does not work? If so, that's really "interesting".

lib/devos/mkProfileAttrs.nix
blaggacao: `disabledModule` is less brittle when working with keys. In a sharing model, modules can also come from external devos flakes.

Equality of store paths is cheap even in those scenarios. Equality of relative strings sets us up for trouble.

blaggacao:
Can you check if it works with just declaring relative paths? (`./hosts`, etc.)

Otherwise this changes is already really enjoyable to see.

flake.nix
Pacman99: Yup, I dealt with this in the `mkFlake` PR by converting everything to string using `apply` keys. But I don't have that option here. I think its because of `builtins.readDir` reacting weirdly to paths. But overall I think I've found when working with store paths, everything is just easier if their strings.

Pacman99:
> Can you check if it works with just declaring relative paths? (./hosts, etc.)

Heres the error I get:
```sh
error: access to path '/nix/store/i3z291bqqjmsgr4mpv8ys90da7h1nazy-hosts/NixOS.nix' is forbidden in restricted mode
```
I have no clue whats happening, because that store path that it is talking about actually exists and there is a README and NixOS.nix in there. Which means at some point some function ended up getting nix to create a new store path for the directory it was reading with the necessary files, and then error itself out because it doesn't ave access to those files.

lib/devos/mkHosts.nix
```diff
@@ -80,7 +70,7 @@ let
     let
       local = {
         require = [
-          "${toString ./.}/${hostName}.nix"
+          "${dir}/${hostName}.nix"
```
blaggacao, line 73: Does the error persist if you convert this back to a path?

blaggacao (commented):
Maybe this could be the culprit...

lib/devos/mkHosts.nix
Pacman99: ohhh good catch, let me try

Pacman99: actually I think its the opposite, we need to convert dir to a string

Pacman99: well I guess that comes back to the original issue, we have to specifically convert everything to a string, or pass them as a a string. I think its an issue with `builtins.readDir`

Pacman99:
```suggestion
          "${toString dir}/${hostName}.nix"
```
this *fixes* it

Pacman99:
So the only solution I can think of for the paths issue is change any lib function that uses `builtins.readDir` to convert the dir to a string before passing it to the builtin. I really don't like the solution though.

blaggacao:
> I really don't like the solution though.

If we call it an "anti corruption layer" against `nix` quircks, would that appease?

I think working with paths is always a better option, simply because it is a richer type to mean the same.

Pacman99:
I would actually argue the opposite, specially for store paths. All of nixpkgs passes around store paths as strings. And I think nix in general treats paths as a shorthand to represent a string. So paths actually have more chance for corruption since they can be relative or absolute and can cause issues with builtin functions. String paths on the other hand are always absolute and they universally work on all builtin functions. While I would like users to be able to pass in paths, I think of it more of a convenience and the mkFlake PR does allow this and solves the issue by converting to string immediately.

Pacman99:
But for this PR since we are still in the stage of devos where users aren't necessarily meant to edit their flake.nix, It think the current solution is best to just pass the paths as strings for profiles/hosts.

blaggacao:
LGTM! Let's address this separately and get this rather merged.

Strange thing that nixpkgs prefers strings over paths. At the very least it's bad semantics and an absent anti corruption fuse somwhere deep down.

blaggacao (approved):
LGTM (from cell phone)

Pacman99:
huh `core` is being imported twice in profilesTest

nrdxp:
Is it the same issue from #136? Essentially we fixed it by insuring that all profiles in mkProfilesAttrs a just a path to the profile, since the nix module system seems to compare the paths and remove redundant entries from imports, but it doesn't do the same for a module function.

blaggacao:
:+1: for filtering out `core` from `allProfiles`.

lib/devos/mkProfileAttrs.nix
nrdxp: I believe this is the cause of the BORS failure. For whatever reason, the core in `mkHosts` and this the one imported here have different hashes, meaning toPath must have changed the path somehow.

Pacman99:
Yeah I am a little confused, it should have been able to filter it out, relative and absolute paths compare well
Also random idea, maybe core could be a module thats enabled by default instead. So that way users can customize what they want out of core. something like `devos.core.enable`.

Pacman99:
bors try

nrdxp:
Since this PR also introduces new lib functions. A minimal test case could be in order, to ensure corner cases are worked out, or at least documented before merge. I am gonna try to help out with this later tonight when I have more time.

Pacman99:
> Since this PR also introduces new lib functions. A minimal test case is in order to ensure corner cases are worked out, or at least documented before merge. I am gonna try to help out with this later tonight when I have more time.

That would be great! I'm not sure how to write tests for lib functions, so any demonstration would be helpful.

lib/devos/mkProfileAttrs.nix
nrdxp: This seems to work as just `"${dir}/${n}"`, had you tried that before `toPath`?

nrdxp:
Sorry for force push, meant to send it to divnix branch but got pushed to yours by mistake. I didn't actually change your commits, only added and then removed one, so it should still be the same. Just experimenting with the CI issue.

Pacman99:
#223 would fix the CI issue. But that requires a consensus to actually move core to modules, which is a fairly big change - not code wise but api wise, since historically core has always been a profile.

nrdxp:
Pushed to commits that fixed CI, and removed the call to `builtins.toPath`. Will try to work up some tests for these funcitons now.

nrdxp:
bors try

lib/devos/mkProfileAttrs.nix
Pacman99: No I know that works. The reason this has to be passed as a path is for the `disabledModules` section of `isoConfig`. If you look here: https://github.com/NixOS/nixpkgs/blob/1dfe690209af14a91b32dbf2f379ca02d050e1a1/lib/modules.nix#L241

Any string passed to disabledModules is always appended to `modulesPath`. But if you pass it as a path, then it works.

lib/devos/mkHosts.nix
```diff
@@ -21,10 +12,9 @@ let
   ];
 
   modules = {
-    core = ../profiles/core;
+    core = "${self}/profiles/core";
```
Pacman99, line 15: Ohh I see the problem. I will probably have to update this file then in #218, because I think it should be possible for a user of `mkFlake` to safely delete core. Or #223 would work too.

lib/devos/mkProfileAttrs.nix
Pacman99: Ohh I just read your commit message, I guess `toPath` doesn't help either.

Pacman99: I think this PR will just break isoConfig then. We will need to find a better fix for that.

nrdxp: Ah, now I see, so we could probably just `map (x: /. + x)` to fix isoConfig

lib/devos/mkHosts.nix
blaggacao: Interesting! So the rational is that this also fixes the issue with filtering core from isoConfig?

It would be interesting to know, if store path canonicalization lre-processing on path types actually prevents such unintended identity (hash) mismatches.

But if this doesn't break isoConfig, I'm fine with also removing `/. + ...` further down, for now.

Pacman99:
added a test for suites and profiles. Not sure how to do one for mkHosts, since that results in a `nixosSystem` which is really large, so writing the `expected` for that might not be possible..

nrdxp:
`mkHosts` should be fine without testing, since its results are technically tested when the test VM is booted in `profilesTest`. Of course I would like to make some more elaborate tests in the future, which will all rely on the logic in `mkHosts` to provide the system environment. 

An example that comes to mind is shutdown hang (I hate it when systemd hangs on shutdown, I'd at least like to know if an update to nixpkgs is gonna cause my system to do it).

Pacman99:
That makes sense, then hopefully the `testSuites` I added is enough.
I cleaned up git history, and since mkHosts test isn't needed, this ready for review.

lib/devos/devosSystem.nix
```diff
@@ -18,7 +18,10 @@ lib.nixosSystem (args // {
             ({ config, suites, ... }: {
 
               # avoid unwanted systemd service startups
-              disabledModules = lib.remove modules.core suites.allProfiles;
+              # all strings in disabledModules get appended to modulesPath
+              # so convert each to list which is not a string but can be coerced to one
+              disabledModules = map (x: [ x ])
+                (lib.remove modules.core suites.allProfiles);
```
Pacman99, line 24: Best fix I could come up with to ensure the profiles still get disabled since we now pass profiles as strings again.

Pacman99: why this is necessary: https://github.com/NixOS/nixpkgs/blob/1dfe690209af14a91b32dbf2f379ca02d050e1a1/lib/modules.nix#L241

Pacman99: and nix doesn't allow store paths to be appended to relative paths

nrdxp (approved):
bors r+

PR #217, opened 2021-03-26
lib/mkPkgs: don't import external files and get inputs as argument

Pacman99:
Changes mkPkgs to take extern, overrides, and pkgs as an argument. This improves their ability to be used as lib functions.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -89,7 +89,7 @@
             (os.mkPackages { inherit pkgs; });
 
           devShell = import ./shell {
-            inherit self system;
+            inherit self system extern overrides;
           };
         }
       );
--- a/flake.nix
+++ b/flake.nix
@@ -34,8 +34,11 @@
       inherit (lib) os;
 
       extern = import ./extern { inherit inputs; };
+      overrides = import ./overrides;
 
-      multiPkgs = os.mkPkgs;
+      multiPkgs = os.mkPkgs {
+        inherit extern overrides;
+      };
 
       outputs = {
         nixosConfigurations =
--- a/lib/devos/mkPkgs.nix
+++ b/lib/devos/mkPkgs.nix
@@ -1,13 +1,11 @@
-{ lib, dev, nixos, self, ... }:
+{ lib, dev, nixos, self, inputs, ... }:
 
-let inherit (self) inputs;
-in
+{ extern, overrides }:
 (inputs.utils.lib.eachDefaultSystem
   (system:
     let
-      extern = import ../../extern { inherit inputs; };
       overridePkgs = dev.os.pkgImport inputs.override [ ] system;
-      overridesOverlay = (import ../../overrides).packages;
+      overridesOverlay = overrides.packages;
 
       overlays = [
         (overridesOverlay overridePkgs)
--- a/shell/default.nix
+++ b/shell/default.nix
@@ -1,8 +1,12 @@
 { self ? (import ../compat).defaultNix
 , system ? builtins.currentSystem
+, extern ? import ../extern { inherit (self) inputs; }
+, overrides ? import ../overrides
 }:
 let
-  pkgs = (self.lib.os.mkPkgs).${system};
+  pkgs = (self.lib.os.mkPkgs {
+    inherit overrides extern;
+  }).${system};
 
   inherit (pkgs) lib;
 


shell/default.nix
```diff
@@ -1,9 +1,8 @@
 { self ? (import ../compat).defaultNix
 , system ? builtins.currentSystem
+, pkgs ? import <nixpkgs> {}
```
nrdxp, line 3: This may not be such a great idea, as there is no guarantee the NIX_PATH is gonna contain a `nixFlakes` version that works with the _flk.sh_ script.

Pacman99: Ok I can remove that change, it's pretty unnecessary.

Pacman99: removed that change.

lib/devos/mkPkgs.nix
```diff
@@ -1,13 +1,11 @@
-{ lib, dev, nixos, self, ... }:
+{ lib, dev, nixos, self, inputs, ... }:
 
-let inherit (self) inputs;
-in
+{ extern, overrides }:
```
nrdxp, line 3: wouldn't this have to come first?

Pacman99: Can you elaborate? first as in before what?

nrdxp: As in the first arguement, before `{ lib, dev, nixos, self, inputs, ... }:`

Pacman99: really? that would fail, the first line has to be the lib's dependencies and the second line is the actual function arguments. Not sure if I'm just missing something.

nrdxp: Oops, looks like I forgot that the first arg is passed automatically during lib creation. My bad.

nrdxp (approved):
This looks good. Just want to ask if it's all finished before merging.
bors try

nrdxp:
Ah, BORS failed since you added a second argument to mkPkgs, so the call from _shell/default.nix_ needs to be updated with the new args as well.

Pacman99:
Right, now I remember why I added the pkgs there in the first place. Uhh I'm not sure how to do this. 
Shell could manually call extern, but it doesn't have access to the right inputs. Well it does right now, but in the future `self` might not actually refer to devos, which is why I switched mkPkgs to get inputs from lib so it has all of them. The other option is pass extern and overrides to shell as we do to pkgs, but that would break compat.

shell/default.nix
```diff
@@ -1,8 +1,12 @@
 { self ? (import ../compat).defaultNix
 , system ? builtins.currentSystem
+, extern ? import ../extern { inherit (self) inputs; }
+, overrides ? import ../overrides;
 }:
 let
-  pkgs = (self.lib.os.mkPkgs).${system};
+  pkgs = (self.lib.os.mkPkgs {
+    inherit overrides extern;
+  }).${system};
```
Pacman99, line 9: I think this should work. The only issue is If shell is ever called with `nix-shell` then it might not have all the right inputs to pass to extern. We could always try to fix this later, not sure.

flake.nix
```diff
@@ -86,7 +89,7 @@
             (os.mkPackages { inherit pkgs; });
 
           devShell = import ./shell {
-            inherit self system;
+            inherit self system extern overrides;
```
nrdxp, line 92: overrides is not in scope here, causing an error.

nrdxp (changes requested):
Looks like it's almost ready.

flake.nix
Pacman99: ok right fixed, I think I forgot to run a `nix flake check` before pushing this change.

shell/default.nix
nrdxp: `nix-shell` actually works fine in this case, after testing on my end.

nrdxp (approved):
bors r+

shell/default.nix
Pacman99: It does now I'm more talking about future when mkFlake calls shell. The self that it gets might be the flake thats calling it. And if you notice in the `mkFlake` PR I pass both the current self's inputs and devos's inputs to extern. That won't happen here, so if a user's extern was, for example, referring to deploy-rs, but they didn't include it in their own flake, shell would fail.

Issue #214, opened 2021-03-26
devos-as-library


blaggacao:
# Proposed API (v0.2)
```nix
# flake.nix
{
  inputs.devos.path = ./lib;
  # inputs.devos.url = "github.com:divnix/devos?dir=lib";
  outputs = inputs@{self, devos}: let name = "mydevos"; in
  devos.mkFlake {
    inherit self inputs name;
    hosts = ./hosts; # or attrset, usable also: devos.lib.attrsFromPath
    profiles = ./profiles; # or attrset
    modules = [ ]; # or path
    suites = { }; # or path
    packages = name: final: prev: { }; # name spacing external pkgs sharing | or path
    "..." # v0.2 place holder ;-)
  }
  //
  {
    # extend with custom stuff
    hmConfigurations.some = devos.mkHmConfiguration {};
    nixosConfigurations = lib.mapAttrs (n: v: {name = n + "-staging"; value = v;}) (devos.lib.mkAttributesFromPath ./staging);
    devshellModules = {}; # prototyping / showcasing future ideas for devos-as-library
  };
}
```

# Rationale

(tbd &mdash; will redact later)

---

### Related / Affected / Solvable
- #36
- #91
- #121
- #151
- #166
- #179
- #183

blaggacao:
To complement the collaboration model to somebody who wants to extend devos lib itself, we can recommend to do it as a patch / PR in the spirit of https://github.com/NixOS/nix/issues/3920

Pacman99:
I like this a lot :+1: as I mentioned earlier, I would prefer it if we first actually create the library function for mkFlake and have that not implement a new api design and just implement the current api. And then once we finalize the devos api, we can do another PR to update mkFlake/evalArgs.

Pacman99:
On the proposed api itself. I was considering allowing paths, but then decided against it. It makes it harder to type check things within the module system and when we don't allow paths, we can add some other features. For example the current #206, actually lets you drop certain attributes within files like suites.userSuites. Because the suites apply key will just append an empty attrset for anything that you don't include. And that feature is much harder to implement if we allow paths.

So essentially you have to do an `import ./suites` if its not in the default location.

nrdxp:
> We should also decide on where to use auto-import magic. I'm fine with it as its used right now. hosts are auto-imported and have clear rules for how they are imported. Profiles are technically auto-imported, but because of how suites work, you still explicitly have to select suites, so I think thats ok.

[...]

PR #218, opened 2021-03-26
lib: init and use mkFlake and evalFlakeArgs

Pacman99:
Add general lib functions to create a devos flake. These functions do not enforce a file system structure, but just provide an api to create a flake. 
The options that have types do not allow paths, see [this comment](https://github.com/divnix/devos/issues/214#issuecomment-808357834) for my reasoning. 
Module system allows for type checking and storing all data about the flake within one comprehensive evaluated module. It also allows for adding defaults, so folders and files are no longer required. You can safely delete the `userSuites` or just not pass modules if you aren't using those features, and devos won't error out like it does now.

Then in devos's flake.nix we now call mkFlake and pass the arguments to enforce this template's file system structure.

I mentioned this in other places, but I do not want to change any of devos's api in this PR. So for example @blaggacao has brought up that extern might be somewhat necessary and could be changed. But I'm still going to keep extern in this PR since thats how devos works right now. Then after the result of #214 and/or #152 we can update mkFlake and devos's flake.nix appropriately in separate PR's.

Thank you @blaggacao for the great ideas! Credit goes to him for the idea of making mkFlake more general. It took a bit for me to understand, but this is way better.

### DIFF
--- a/doc/start/index.md
+++ b/doc/start/index.md
@@ -33,6 +33,7 @@ In addition, the [binary cache](../../cachix) is added for faster deployment.
 >   upstream changes.
 
 ## Next Steps:
+- [Use Devos as a library!](./mkflake.md)
 - [Make installable ISO](./iso.md)
 - [Bootstrap Host](./bootstrapping.md)
 - [Already on NixOS](./from-nixos.md)
--- a/flake.nix
+++ b/flake.nix
@@ -31,8 +31,8 @@
     outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
       let
         lib = import ./lib { inherit self nixos inputs; };
-
-        out = lib.mkFlake {
+      in
+        lib.mkFlake {
           inherit self;
           hosts = ./hosts;
           packages = import ./pkgs;
@@ -44,18 +44,16 @@
           userProfiles = ./users/profiles;
           modules = import ./modules/module-list.nix;
           userModules = import ./users/modules/module-list.nix;
+        } // {
+          defaultTemplate = self.templates.flk;
+          templates.flk.path = ./.;
+          templates.flk.description = "flk template";
+          templates.mkflake.path =
+            let
+              excludes = [ "lib" "tests" "cachix" "nix" "theme" ".github" "bors.toml" "cachix.nix" ];
+              filter = path: type: ! builtins.elem (baseNameOf path) excludes;
+            in
+              builtins.filterSource filter ./.;
+          templates.mkflake.description = "template with necessary folders for mkFlake usage";
         };
-
-      in nixos.lib.recursiveUpdate out {
-        defaultTemplate = self.templates.flk;
-        templates.flk.path = ./.;
-        templates.flk.description = "flk template";
-        templates.mkflake.path =
-          let
-            excludes = [ "lib" "tests" "cachix" "nix" "theme" ".github" "bors.toml" "cachix.nix" ];
-            filter = path: type: ! builtins.elem (baseNameOf path) excludes;
-          in
-            builtins.filterSource filter ./.;
-        templates.mkflake.description = "template with necessary folders for mkFlake usage";
-      };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -28,79 +28,34 @@
       pkgs.inputs.nixpkgs.follows = "nixos";
     };
 
-  outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
-    let
-      inherit (self) lib;
-      inherit (lib) os;
+    outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
+      let
+        lib = import ./lib { inherit self nixos inputs; };
 
-      extern = import ./extern { inherit inputs; };
-      overrides = import ./overrides;
-
-      multiPkgs = os.mkPkgs {
-        inherit extern overrides;
-      };
-
-      suites = os.mkSuites {
-        suites = import ./suites;
-        users = os.mkProfileAttrs "${self}/users";
-        profiles = os.mkProfileAttrs "${self}/profiles";
-        userProfiles = os.mkProfileAttrs "${self}/users/profiles";
-      };
-
-      outputs = {
-        nixosConfigurations = os.mkHosts {
-          dir = "${self}/hosts";
+        out = lib.mkFlake {
+          inherit self;
+          hosts = ./hosts;
+          packages = import ./pkgs;
+          suites = import ./suites;
+          extern = import ./extern;
           overrides = import ./overrides;
-          inherit multiPkgs suites extern;
+          overlays = ./overlays;
+          profiles = ./profiles;
+          userProfiles = ./users/profiles;
+          modules = import ./modules/module-list.nix;
+          userModules = import ./users/modules/module-list.nix;
         };
 
-        homeConfigurations = os.mkHomeConfigurations;
-
-        nixosModules =
-          let moduleList = import ./modules/module-list.nix;
-          in lib.pathsToImportedAttrs moduleList;
-
-        homeModules =
-          let moduleList = import ./users/modules/module-list.nix;
-          in lib.pathsToImportedAttrs moduleList;
-
-        overlay = import ./pkgs;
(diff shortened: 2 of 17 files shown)


Pacman99:
Moved templates and devosOptionsDoc out of `mkFlake`, since they are devos specific. devosOptionsDoc is added to pkgs, I think that would be the best place to add it, following the devos api thats just been defined.

Pacman99:
Dependent PR's:
 - #217 
 - #216 
 - #219 
 - #215

nrdxp:
I want to filter packages by system, but I don't think we actually need the flattentree funcitonality anhow, so perhaps we could make it simpler and invert the logic, so that if a system isn't explicitly specified, instead of being filtered out as it is now, it is automatically included in all systems. This encourages packages that should be systemed to do so, or CI won't pass, while allowing new tests and the like to automatically run on any system.

lib/mkFlake/evalArgs.nix
```diff
@@ -0,0 +1,163 @@
...
+      options = with types; {
+        self = mkOption {
+          type = addCheck attrs nixos.lib.isStorePath;
+          description = "The flake to create the devos outputs for";
+        };
+        hosts = mkOption {
+          type = path;
+          default = "${self}/hosts";
+          defaultText = "\${self}/hosts";
+          apply = toString;
```
Pacman99, line 25: toString is necessary here otherwise for some reason the `mk*` functions create a new store path with just that folder and then nix fails because of an access to a restricted store path. either way its better to pass these paths as strings.

Pacman99:
Added a doc section, 'use devos as a library!' which is an updated version of the 'use-mkdevos' doc from the other PR.

lib/devos/mkHosts.nix
```diff
@@ -1,6 +1,6 @@
 { lib, dev, nixos, inputs, self, ... }:
 
-{ dir, extern, suites, overrides, multiPkgs, ... }:
+{ dir, devos, extern, suites, overrides, multiPkgs, ... }:
```
Pacman99, line 3: So we users of mkflake can safely delete the core profile or cachix module.

lib/mkFlake/default.nix
```diff
@@ -0,0 +1,64 @@
...
+    let
+      pkgs = multiPkgs.${system};
+      # all packages that are defined in ./pkgs
+      legacyPackages = os.mkPackages { inherit pkgs; };
+    in
+    {
+      checks =
+        let
+          tests = nixos.lib.optionalAttrs (system == "x86_64-linux")
+            (import "${devos}/tests" { inherit pkgs; self = devos; });
```
Pacman99, line 48: I think passing `devos` as self is optimal for now. We know devos will always have `self.nixosConfigurations.NixOS`, while end user might delete the default hosts. And we still have test coverage of `self` with deploy-rs checks. But in the future I think it would be good to pass both `self` and `devos`. And have a mix of testing devos's hosts, and the users hosts/profiles. But lets save for a future PR.

Pacman99:
> What's left to do here, before review?

I just went through a few cleanups today and improved the option descriptions. This should be ready for review. I don't plan on adding anything else in the PR.

lib/mkFlake/evalArgs.nix
```diff
@@ -0,0 +1,182 @@
...
+    {
+      options = with types; {
+        self = mkOption {
+          type = addCheck attrs nixos.lib.isStorePath;
+          description = "The flake to create the devos outputs for";
+        };
+        hosts = mkOption {
+          type = path;
+          default = "${self}/hosts";
+          defaultText = "\${self}/hosts";
```
Pacman99, line 26: Defaults for all path options do end up adding some file system structure. I think its ok since it is a path option. But just wanted to note that.

lib/devos/mkProfileAttrs.nix
```diff
@@ -16,7 +16,7 @@ let mkProfileAttrs =
   let
     imports =
       let
-        files = builtins.readDir dir;
+        files = lib.optionalAttrs (builtins.pathExists dir) (builtins.readDir dir);
```
Pacman99, line 19: To add more safety to the profile options

Pacman99:
bors try

Pacman99:
running into the double importing `core` issue again. I thought `${devos}/profiles/core` would be the same core in `allProfiles`. I'm confused as to why thats not the case, since `devos == self`

nrdxp:
Oh wow, it's even stranger this time, since the hashes actually match exactly as well. I'll have to experiment a bit and see if I can find a workaround/solution.

blaggacao (commented):
Looks like.my feedback didn't make it through once again. No worries, don't mean to block anything. Take what you find useful...

(Sorry for the procediral complications on my end)

pkgs/default.nix
```diff
@@ -1 +1,3 @@
-final: prev: { }
+final: prev: {
```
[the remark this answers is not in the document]
Pacman99: good point! I'm going to leave it with this method for now - since it doesn't require too much refactoring, but definitely something to consider in the future.

flake.nix
```diff
@@ -28,72 +28,35 @@
...
-          in lib.pathsToImportedAttrs moduleList;
-
-        overlay = import ./pkgs;
-        overlays = lib.pathsToImportedAttrs (lib.pathsIn ./overlays);
-
-        lib = import ./lib { inherit nixos self inputs; };
-
+    outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
+      let
+        lib = import ./lib { inherit self nixos inputs; };
```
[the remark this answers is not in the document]
Pacman99: Overall this is a good idea, but it would require a bit of refactoring that I would rather not include in this PR.

lib/mkFlake/default.nix
```diff
@@ -0,0 +1,64 @@
...
+  inherit (self) lib;
+  inherit (lib) os;
+
+  inherit (inputs) utils deploy;
+
+  cfg = (lib.evalFlakeArgs { inherit args; }).config;
+
+  multiPkgs = os.mkPkgs { inherit (cfg) extern overrides; };
+
+  outputs = {
```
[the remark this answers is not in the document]
Pacman99: I prefer `outputs` and `systemOutputs` just because its really clear as to what each is. And less mental hoops to jump through when dealing with `mkFlake`. If it becomes more common, then maybe we could switch.

doc/start/index.md
```diff
@@ -36,6 +36,7 @@ In addition, the [binary cache](../../cachix) is added for faster deployment.
 - [Make installable ISO](./iso.md)
 - [Bootstrap Host](./bootstrapping.md)
 - [Already on NixOS](./from-nixos.md)
+- [Use Devos as a library!](./mkflake.md)
```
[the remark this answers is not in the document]
Pacman99: fixed

lib/mkFlake/default.nix
```diff
@@ -0,0 +1,67 @@
...
+
+    homeModules = cfg.userModules;
+
+    overlay = cfg.packages;
+    inherit (cfg) overlays;
+
+    lib = import "${devos}/lib" {
+      inherit self nixos;
+      inputs = inputs // self.inputs;
+    };
```
Pacman99, line 36: Just added this, so we can also extend devos api to flake inputs. Creates opportunity to do things based on what input exists, allowing integrations to become optional.

Pacman99: for example, `// optionalAttrs (inputs ? home) { home-manager = ... }`

Pacman99:
bors try

Pacman99:
More dependent PR's:
 - [] #229, this should help with path issue, but its not necessary
 - [] #228 
 
 With those PR's, the current "black magic" will no longer be necessary. `lib` only needs to be initialized once and self will just be self across lib, the flake that is calling each lib function. And `devos` isn't necessary. 
 The only remaining issues is tests and shell. I think we can just expect users of `mkFlake` to leave those folders for now. In future, shell should prob be moved to lib and only the `compat` part should be in the root folder. And we could add a `tests` argument for `mkFlake` - that will require more tests rework, so not planning to include here.

Pacman99:
mostly implemented in #231

PR #220, opened 2021-03-26
Drop flattenTreeSystem and use custom function for filtering packages

Pacman99:
I don't think we should flatten the system because if a user doesn't make a package a derivation in pkgs/default.nix we should trust that there is a reason for doing so. So instead this drops the flattenTreeSystem reference(and switches to flake-utils master branch) and replaces its usage with a custom function `filterPackages`. This function filter all packages that match three conditions;
 - is a derivation
 - not broken
 - system is supported

In that order as to not cause errors when trying to reference non-derivation meta attributes.

And then also just dump *all* packages into legacy packages, so everything else is still accessible. I was considering removing the packages that are already in the packages output in legacyPackages, but I don't think its necessary since nix looks to the packages output first.

### DIFF
--- a/lib/attrs.nix
+++ b/lib/attrs.nix
@@ -26,4 +26,13 @@ rec {
     });
 
   concatAttrs = lib.fold (attr: sum: lib.recursiveUpdate sum attr) { };
+
+  # Filter out packages that support given system and follow flake check requirements
+  filterPackages = system: packages:
+    let
+      # Everything that nix flake check requires for the packages output
+      filter = (n: v: with v; let platforms = meta.hydraPlatforms or meta.platforms or [ ]; in
+        lib.isDerivation v && !meta.broken && builtins.elem system platforms);
+    in
+      lib.filterAttrs filter packages;
 }
--- a/flake.lock
+++ b/flake.lock
@@ -260,16 +260,15 @@
     },
     "utils": {
       "locked": {
-        "lastModified": 1608663846,
-        "narHash": "sha256-YOAXeoNLW0IubQ6AZLGuSj5kAD5n4DYXLYReLtyds1A=",
+        "lastModified": 1614513358,
+        "narHash": "sha256-LakhOx3S1dRjnh0b5Dg3mbZyH0ToC9I8Y2wKSkBaTzU=",
         "owner": "numtide",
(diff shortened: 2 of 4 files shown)


lib/attrs.nix
```diff
@@ -26,4 +26,13 @@ rec {
     });
 
   concatAttrs = lib.fold (attr: sum: lib.recursiveUpdate sum attr) { };
+
+  # Filter out packages that support given system and follow flake check requirements
+  filterPackages = system: packages:
+    let
+      # Everything that nix flake check requires for the packages output
+      filter = (n: v: lib.isDerivation v && !v.meta.broken &&
+        builtins.elem system v.meta.platforms);
```
nrdxp, line 35: May want to add a let binding to `filter` such as:
```nix
let platforms = meta.hydraPlatforms or meta.platforms or [ ]; in # ...
builtins.elem system platforms
```
Code comes directly from `flattenTreeSystem` itself since some packages define only `hydraPlatforms`, while others define only `platforms`. The empty list ensure evaluation won't fail if neither exist.

nrdxp (changes requested):

lib/attrs.nix
Pacman99: just updated with suggestion

nrdxp (changes requested):
The code changes are good, but for the sake of a prettier git log, the first line of a commit message should be kept under 50 chars, while subsequent lines should remain under 72 chars

Pacman99:
> The code changes are good, but for the sake of a prettier git log, the first line of a commit message should be kept under 50 chars, while subsequent lines should remain under 72 chars

fixed! I really should stop using `git commit -m`

nrdxp (approved):
It's all good. I've made the same mistake. Good work!
bors r+

blaggacao:
What is this mainly motivated by (for the record)?

Is this motivated by the issues we had with flake checks & multiplatform?

nrdxp:
some [context](https://github.com/divnix/devos/pull/218#issuecomment-808487785)

Also `flattenTreeSystem` never got merged into flake-utils and was instead closed, so if the branch ever got deleted users might be  pretty upset as well.

PR #222, opened 2021-03-27
flake: Fix #211 remove ImportedAttr of moduleList

GTrunSec:


nrdxp:
If I understand this patch correctly, it breaks exporting of individual, user defined `hmModules` which may not be a good idea. Essentially home-manager modules are a mirror of `nixosModules`, that is an attributeset of modules. If you are having trouble importing a user defined module just be sure to reference it by filename.

@Pacman99, could we get your input here as well, in case I'm missing something.

Pacman99:
I think this will one, break `nix flake check` - haven't tested that. Flake modules are supposed to be an attrset of the name to file/module. As I pointed out in the issue, the better solution would probably be to fix the fact that modules can't be in folders.
But I'm not sure if thats the issue @GTrunSec had, since they were able to access their first module.

GTrunSec:
> I think this will break nix flake check 
yes, I did a bad idea here sorry.

Pacman99:
> I think this will break nix flake check
yes, I did a bad idea here sorry.

No its ok! Its good that you pointed this out and now we know where the issue is. We just have to improve upon your solution to come up with something that solves the issue without breaking other things.

PR #225, opened 2021-03-30
lib: only readDir if path exists with safeReadDir

Pacman99:
After doing this for the second time, I realized it might be good to make a lib function for it and do it across lib.
Create a function called `safeReadDir` that only uses `builtins.readDir` if the path exists. With `optionalAttrs` any function that relies on the output won't fail since they still get an empty attrset.
Then replace all uses of `readDir` with the safe version.


nrdxp (approved):
LGTM

bors r+

PR #226, opened 2021-04-01
Check if options exist before setting them

Pacman99:
mkHosts is not very portable since it relies on nixos options. Which means in the future it can't be used on nix-darwin or robotnix. So here I just wrapped every option thats set to first check if it exists.

I would also like to set the precedent for future integrations added into devos to be made optional with similar methods:
 - Don't add a package unless it exists, ex: `pkgs ? deploy-rs`
 - Don't set an option unless it exists, ex: `options ? home-manager`. This one is tricky, it can't be done with `mkIf`. You have to use a combination of `mkMerge` and `optionalAttrs`, the former is usually necessary when your setting other things and the same module and you want to set an option that might not exist. See changes in devosSystem and mkHosts for examples.
 - Don't export an output that depends on a specific input, unless the input exists: ex: `inputs ? deploy`
 - Don't test relevant outputs unless they exist, ex: `self ? homeConfigurations`. Not sure what to do about this one though, most tests are tightly integrated with devos.

### DIFF
--- a/lib/devos/mkHosts.nix
+++ b/lib/devos/mkHosts.nix
@@ -23,42 +23,54 @@ let
           modules;
       };
 
-    global = { config, pkgs, ... }: {
-      home-manager = {
-        useGlobalPkgs = true;
-        useUserPackages = true;
+    global = { config, options, pkgs, ... }:
+      with lib; lib.mkMerge [
+        (optionalAttrs (options ? home-manager) {
+          home-manager = {
+            useGlobalPkgs = true;
+            useUserPackages = true;
 
-        extraSpecialArgs = extern.userSpecialArgs // { suites = suites.user; };
-        sharedModules = extern.userModules ++ (builtins.attrValues self.homeModules);
-      };
-      users.mutableUsers = lib.mkDefault false;
-
-      hardware.enableRedistributableFirmware = lib.mkDefault true;
-
-      nix.nixPath = [
-        "nixpkgs=${nixos}"
-        "nixos-config=${self}/compat/nixos"
-        "home-manager=${inputs.home}"
+            extraSpecialArgs = extern.userSpecialArgs // { suites = suites.user; };
+            sharedModules = extern.userModules ++ (builtins.attrValues self.homeModules);
+          };
+        })
+        (optionalAttrs (options ? hardware.enableRedistributableFirmware) {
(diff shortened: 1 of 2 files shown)


lib/devos/mkHosts.nix
```diff
@@ -24,39 +24,41 @@ let
...
-    };
+        nix.extraOptions = ''
+          experimental-features = ${lib.concatStringsSep " "
+            experimentalFeatures
+          }
+        '';
+
+        system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+      }
+      (lib.optionalAttrs (options ? home-manager) {
```
Pacman99, line 52: This can't just be `// optionalAttrs ...`, that causes an infinite recursion error.

Pacman99: got `mkMerge` strategy from https://discourse.nixos.org/t/how-to-check-is-an-option-available/2919/2

blaggacao:
Does process on https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/18 feed any information backnon this?

Pacman99:
I actually did that PR with knowledge I gained from working on this.

blaggacao:
Just realized, once we implement config generators in a `lib`, that lib ought to bring its own dependencies. The use of deploy-rs or home-manager will then be optional be the way the user interfaces with that lib's api.

Pacman99:
I think the code in mkHosts, is relevant, where we set the home-manager defaults if the option exists. And there are a couple other things, that are useful, for example deploy might not be in the users packages, so we don't want to add it to shell, unless it exists.

nrdxp:
Is this still an issue/needed?

blaggacao:
We should put it on-hold until we have devos lib and clarity of which delendency can be expected where, we might esle get dead code out of this while refactoring.

Pacman99:
Yeah most of this is unnecessary with the move to pkgs-lib. I think we have now forced lib to depend on deploy-rs.
But I do want to keep this PR for the goal of making home-manager to be optional. I think my other goal for this PR was to have a demonstration of how integrations can be optional to encourage it in the future.

Pacman99:
planning to rework how mkHosts works, so I might include something similar in that PR. But we'll see where it goes.

PR #228, opened 2021-04-03
drop lib's self dep and pass data to each function

Pacman99:
Making lib dependent on devos, was not the greatest idea since it prevents others from using the devos lib. And theres quite a few nice lib functions that people might like that they can't use right now: mkPkgs, mkPackages, mkHomeConfigurations.

This gives the specific data to each lib function that that function needs.

### DIFF
--- a/lib/devos/mkPkgs.nix
+++ b/lib/devos/mkPkgs.nix
@@ -1,13 +1,14 @@
-{ lib, dev, nixos, self, inputs, ... }:
+{ lib, dev, nixos, inputs, ... }:
 
-{ extern, overrides }:
+# main overlay and extra overlays
+{ overlay, overlays, extern, overrides }:
 (inputs.utils.lib.eachDefaultSystem
   (system:
     let
       overridePkgs = dev.os.pkgImport inputs.override [ ] system;
       overridesOverlay = overrides.packages;
 
-      overlays = [
+      allOverlays = [
         (final: prev: {
           lib = prev.lib.extend (lfinal: lprev: {
             inherit dev;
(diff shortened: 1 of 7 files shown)


blaggacao:
I think we should rather go the contrary route and only pass deps that are strictly necesary down to second and third order functions.

If a humongous, all-encompassing `self` needs to be passed through more than one hop, we should rather ask ourselves if it's time for refactoring and untangling the code, before loading it with more functionality.

I think revising/refactoring is in first order, before we can / should implement this feature. Otherwise, we risk in getting this wrong and I don't know the costs of cleanup vs the benefits of the feature.

Pacman99:
all this PR is doing is reverting an old change. So with or without this, cleanup is necessary. But with this, we can get #218 merged, which makes lib a lot easier to work with, for me at least. And will allow out of tree mkflake usage.
I think the benefits far outweigh the slight harms.

lib/devos/mkPkgs.nix
```diff
@@ -1,13 +1,14 @@
-{ lib, dev, nixos, self, inputs, ... }:
+{ lib, dev, nixos, inputs, ... }:
 
-{ extern, overrides }:
+# main overlay and extra overlays
+{ overlay, overlays, extern, overrides }:
```
Pacman99, line 4: I think this is the best way to do this, its hard to combine them all into one argument even though they are related. I chose this method as to not add more implementation logic to combine overlays.

blaggacao (approved):
LGTM, as I understand this PR is a cleanup to limit internal api surface! Nice!

Pacman99:
Converted to draft, so we only merge once 0.9 is in

Pacman99:
I think it might be good to include this in 0.9(contrary to what I said before), but its not necessary.

nrdxp:
bors try

blaggacao:
Superseded by #251

PR #229, opened 2021-04-03
make core an unecessary profile and add to suites

Pacman99:
Prevents mkHosts subverting standard devos api to import core and add
all necessary core features to mkHosts, so core can be safely deleted in
suites

### DIFF
--- a/lib/devos/mkHosts.nix
+++ b/lib/devos/mkHosts.nix
@@ -12,7 +12,6 @@ let
   ];
 
   modules = {
-    core = "${self}/profiles/core";
     modOverrides = { config, overrideModulesPath, ... }:
       let
         inherit (overrides) modules disabledModules;
@@ -24,7 +23,7 @@ let
           modules;
       };
 
-    global = { config, ... }: {
+    global = { config, pkgs, ... }: {
       home-manager = {
         useGlobalPkgs = true;
         useUserPackages = true;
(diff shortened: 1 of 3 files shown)


lib/devos/mkHosts.nix
```diff
@@ -49,6 +49,11 @@ let
         override.flake = inputs.override;
       };
 
+      nix.allowedUsers = [ "@wheel" ];
+      nix.trustedUsers = [ "root" "@wheel" ];
```
blaggacao, line 53: 1. Is there a guarantee for root to exist? 2. Does it matter?
3. Does this need to be here?

blaggacao (commented):
Please falsify my suspicions, here...

lib/devos/mkHosts.nix
Pacman99: I believe root is necessary for a nixos system, I could be wrong.
But that could go in the core profile, I'm not against that. devos would work fine without those lines. 
The separation I did was fairly opinionated, and I've always had those lines in my hosts, so I thought it would make sense to include in mkHosts.

blaggacao: IIRC, @nrdxp did not have root in his config?

Pacman99: Ok I'll just put this back in the core profile. It is definitely not needed, I guess I didn't think there would be a reason to not include it. But you have proved me wrong :).

nrdxp (approved):
bors r+

nrdxp (approved):
bors r+

nrdxp (approved):
bors r+

PR #231, opened 2021-04-06
Move flake implementation logic to lib

Pacman99:
This is a simpler version of #218 that moves flake logic to lib and adds a module to evaluate devos. This DOES NOT support out of tree usage, so if you were following any of the previous PR's, the doc sections/examples to use devos as a library will not work. There is work to make a cleaner api and only then will out of tree support work. Until then, this is still useful to simplify devos and clean up a lot of the implementation logic.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,79 +28,26 @@
       pkgs.inputs.nixpkgs.follows = "nixos";
     };
 
-  outputs = inputs@{ deploy, nixos, nur, self, utils, ... }:
-    let
-      inherit (self) lib;
-      inherit (lib) os;
-
-      extern = import ./extern { inherit inputs; };
-      overrides = import ./overrides;
-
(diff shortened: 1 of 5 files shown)


Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
# Future Work
 - switch to use submodules for suites, extern, and overrides
 - make it work out of tree - need a proper api proposal for this
 - generate docs

Not going to do first two yet, rather wait for #232

blaggacao (approved):
LGTM, assuming that this is #218 without docs, let's got!

nrdxp:
I'll merge it. Has anyone actually tested it on their system personally though?

PR #232, opened 2021-04-07
API Discussions

blaggacao:
https://github.com/divnix/devos/issues/214#issuecomment-808763964

&rarr; [rendered](https://github.com/blaggacao/devos/blob/api/specs/API.md)

### DIFF
--- a/API.md
+++ b/API.md
@@ -49,11 +49,11 @@ The devos-lib onion:
 - `divnix/devos` (the lib)
 
 While flake-utils-plus exports a generic [`utils.systemFlake`](https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/51cb739c9c9c2258bc70747eb7bc22975ae244bd/flake.nix#L37) user function 
-and is restricted from including dependencies, devos-lib should export a `devos.systemFlake` kwithout such restrictions and that can include
+and is restricted from including dependencies, devos-lib should export a `devos.systemFlake` without such restrictions and that can include
 additional (eventually third-party mediated) generatros, such as `deploy-rs`, while implementing a strict superset of
 `utils.systemFlake` api and, on the other hand, a `devos.mkFlake` with importer amenities included and a slightly different api for convenience.
 
-Such function would have the following top level API
+### `devos.systemFlake`
 
 ```nix
 # devos.systemFlake
@@ -92,6 +92,36 @@ Such function would have the following top level API
 , ... # passed through / https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/14}
 ```
 
+### `devos.mkFlake`
+
+This is all about importer amentities over `devos.systemFlake`.
+
+For example people should be able to define either paths or functions as arguments.
+Also we might consider special (import !only!) magic like reducing:
+
+```nix
+{
+  nixos = {
+    modules = ./modules;
+    profiles = ./profiles;
+    specialArgs = { }; # https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/13#issuecomment-814512835
+    hosts = ./hosts;
+    suites = ./suites;
+  };
+}
+```
+
+to
+
+```nix
+{
+  nixos = ./os; # with default fallback folder naming scheme
+}
+```
+
+#### Part 1: Passthrough API from `devos.systemFlake`
+
+We should not change the api here, but rather seek upstream amendmend.
 ```nix
 # devos.mkFlake
 {
@@ -106,10 +136,20 @@ Such function would have the following top level API
     patches = [ ];
   }
 , channelsConfig ? { }
+# TODO: two dimensions: 
+# - shared vs. "per-host"
+# - external (consuming other devos flakes) vs internal
 , sharedModules ? [ ] 
 , sharedOverlays ? [ ]
+, ... # passed through
+}
+```
 
-# allow setting thos ine a variety of (polished ways)
+#### Part 2: Devos NixOS/hm/... configuration API
+```nix
+# devos.mkFlake
+{
+# allow setting those in a variety of (polished ways)
 # like suites = { one = {}; two = {};}; instead of ./suites
 , nixos ? {
     modules = ./modules;
@@ -123,6 +163,5 @@ Such function would have the following top level API
     specialArgs = { };
     suites = ./users/suites;
   }
-, ... # passed through
 }
 ```
--- a/API.md
+++ b/API.md
@@ -1,6 +1,6 @@
 # DevOS 2.0
 
-- We want to be a god ecosystem citizen, therefore we want to rebase on [`gytis/flake-utils-plus`](https://github.com/gytis-ivaskevicius/flake-utils-plus).
+- We want to be a good ecosystem citizen, therefore we want to rebase on [`gytis/flake-utils-plus`](https://github.com/gytis-ivaskevicius/flake-utils-plus).
 - `gytis/flake-utils-plus` is designed to be incorporated in upstream flake-utils. As the name suggests, it wraps and extends `flake-utils` for nixos related use cases.
 - `devos-lib` shall be another layer that wraps `flakes-utils-plus` and adds accoustomed devos amentities, such as:
   - automagic importers, for pure convenience, but without prescribing a particular folder layout (that is something for devos templates to achieve).
--- a/API.md
+++ b/API.md
@@ -1,0 +1,164 @@
+# DevOS 2.0
+
+- We want to be a god ecosystem citizen, therefore we want to rebase on [`gytis/flake-utils-plus`](https://github.com/gytis-ivaskevicius/flake-utils-plus).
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -0,0 +1,164 @@
...
+, deployConfigurations ? { } # escape hatch
+, channels ? {
+    input = "...";
+    overlaysBuilder = channels: [ (final: prev: { })];
+    config = { };
+    patches = [ ];
+  }
+, channelsConfig ? { }
+, sharedModules ? [ ] 
+, sharedOverlays ? [ ]
```
blaggacao, line 115: [fup](https://github.com/gytis-ivaskevicius/flake-utils-plus) only differentiates on the axis of shared vs individual.
Since `devos` also aspirates a sharing model, we need to know which modules to re-export for others to consume.

The naming here is not the best, ideally we could find a naming that clearly reflects both dimensions, without becoming an over-complicated matrix.

API.md
```diff
@@ -0,0 +1,164 @@
...
+, nixosUserProfiles ? [ ]
+, nixosSpecialArgs ? [ ]
+, hmModules ? [ ]
+, hmProfiles ? [ ]
+, hmSpecialArgs ? { }
+
+
+  # TODO
+  , suites
+  , hosts
```
blaggacao, line 128: Not sure what we should expect from `suites` and `hosts` importers here...

API.md
```diff
@@ -0,0 +1,168 @@
...
+, deployConfigurations ? { } # escape hatch / https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/14
+, channels ? {
+    input = "...";
+    overlaysBuilder = channels: [ (final: prev: { })]; # TODO: do we want another interface here? Also: how to handle module backports?
+    config = { };
+    patches = [ ];
+  }
+, channelsConfig ? { }
+, sharedModules ? [ ] 
+, sharedOverlays ? [ ]
```
Pacman99, line 154: I would say call `sharedOverlays` just `overlays` and `sharedModules` something like `internalModules`. That makes it clear that those modules are only for simple things and they won't get exported. And theres no need to call it `sharedOverlays` if theres no "unshared overlays".

API.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+```nix
+# devos.mkFlake
+{
+  self
+, defaultSystem ? "x86_64-linux"
+, supportedSystems ? flake-utils.lib.defaultSystems
+, name # or inputs, see: https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/12
```
Pacman99, line 142: If namespacing is no longer a goal, name isn't necessary. `inputs` is also unecessary, because you can just use `self.inputs`.

blaggacao: I think it boils down to how foreign flakes _should_ be consumed. So it's kind of a difficult normative call to make. As I argue in the linked issue, `flake.nix` have a relatively strict output schema. I only might speculate the intention was to also consume flake by them and not the en-globing corresponding input (not `inputs`!) object, itself.

API.md
blaggacao: In theory, we have:
- shared modules (for all hosts)
- non-shared modules (for some hosts)
- shared overlays (for all hosts)
- non shared overlays (for specific hosts)
- but also external shared modules and external host specific modules
- and also external shared and unshared overlays

... _in theory_. We have to make a good cut, I guess.

gytis-ivaskevicius: sharedOverlays vs overlays - calling it `overlays` would mean a name confliction between abstraction and flake itself.
internalModules - I'd say naming could be better than that. Maybe something that actually represents that its auto imported to all the configurations?

Overall, I'm quite a fan of `sharedXyz` since naming is clear, more or less makes sense, and does not conflict with actual flake attributes

API.md
```diff
@@ -0,0 +1,167 @@
+# DevOS 2.0
+
+- We want to be a god ecosystem citizen, therefore we want to rebase on [`gytis/flake-utils-plus`](https://github.com/gytis-ivaskevicius/flake-utils-plus).
+- `gytis/flake-utils-plus` is designed to be incorporated in upstream flake-utils. As the name suggests, it wraps and extends `flake-utils` for nixos related use cases.
+- `devos-lib` shall be another layer that wraps `flakes-utils-plus` and adds accoustomed devos amentities, such as:
+  - automagic importers, for pure convenience, but without prescribing a particular folder layout (that is something for devos templates to achieve).
+  - additional configuration generators, such as for `deploy-rs` or `home-manager` or other future integrations (eg. mobile nixos, etc) that we want to offer as a convenience.
```
gytis-ivaskevicius, line 7: I would be so down with implementing deploy-rs and HM as part of the flake-utils-plus ;)

API.md
```diff
@@ -0,0 +1,167 @@
...
+## Importers
+
+Importers declutter the top level flake. The goal is to have a bunch of lines that read like a very clean table of contents:
+
+```nix
+  concept1 = ./concept1;
+  concept2 = ./concept2;
+  concept3 = { "..." = "..."; };
+  concept4 = [];
+```
```
gytis-ivaskevicius, line 21: Not exactly relevant but having all attributes interpreted like so would be pretty cool:
```nix
{
  concept1 = ./some/directory; # Evaluates to 'concept1 = { fileNameWithoutNixExtension = {...}; }'
  concept2 = ./someFile.nix; # Evaluates to  'concept2 = { "..." = "..."; }'
  concept3 = { "..." = "..."; }; # Evaluates to 'concept3 = { "..." = "..."; }'
  concept4 = [];  # Evaluates to 'concept4 = []'
}
```

API.md
```diff
@@ -0,0 +1,167 @@
...
+, channelsConfig ? { }
+, sharedModules ? [ ] 
+, sharedOverlays ? [ ]
+, nixosSpecialArgs ? { } # https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/13#issuecomment-814512835
+
+, hmConfigurations ? { }
+, hmProfiles ? { NAME = { 
+    modules = [ ];
+    extraArgs = { };
+    # what else makes sense, here?
```
gytis-ivaskevicius, line 87: I believe `system` is still required, also maybe having an option to pick which channel is used would be nice since a lot of people are running core system on stable, HM on unstable channels

API.md
```diff
@@ -0,0 +1,167 @@
...
+    system = "...";
+    channelName = "...";
+    extraArgs = { };
+    modules = [ ];
+  };}
+, channels ? {
+    input = "...";
+    overlaysBuilder = channels: [ (final: prev: { })];
+    config = { };
+    patches = [ ];
```
gytis-ivaskevicius, line 76: At this point I'm like "can we just pay some C++ nerd to properly implement channels patching as part of Nix binary instead" :D

API.md
```diff
@@ -0,0 +1,167 @@
...
+    profiles = ./profiles;
+    specialArgs = { }; # https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/13#issuecomment-814512835
+    hosts = ./hosts;
+    suites = ./suites;
+, hm ? {
+    modules = ./users/modules;
+    profiles = ./users/profiles;
+    specialArgs = { };
+    suites = ./users/suites;
+  }
```
gytis-ivaskevicius, line 165: `hm.modules = ./abc;` is not a valid syntax due to missing profiles/specialArgs/suites. I am not sure if thats intended.

gytis-ivaskevicius (approved):
Okay, first of all, the more I read name `systemFlake` the more i hate it :D

Second, If you guys were to create additional devos.nix file or something like that as part of `fup` for time being - I am fine with it. This might make life easier, especially in a prototyping stage.

API.md
blaggacao: FWOW `simpleFlake` calls it `preOverlays` &mdash; since the order of application matters in overlays.

API.md
blaggacao: Hm, actually that seems like a good venue. We could start by lobbying more heavily and probably Ma29 could lend us a hand.

API.md
blaggacao: The idea here is to take those values at face value & in a first step applying _maybeImporters_ and only in a second step _function instantiators_, the latter passing all nescessary args around.

Does that make sense or did I misread your comment?

API.md
blaggacao: At least deploy-rs currently has a dependency injected of a `deploy.lib` library function, though.

I still regard being dependency-free as a feature. :wink:

API.md
gytis-ivaskevicius: I mean the way default values are implemented - if `hm.modules` is defined - `hm.profiles`/`hm.specialArgs`/`hm.suites` must be manually defined as well

API.md
gytis-ivaskevicius: oh yeah, I am planning to keep it dependency-free :)

API.md
blaggacao: Ah! @Pacman99 in the meanwhile is an expert of validating, crosschecking and defaulting values via the module system.

I think it will be a mix of a:
- coustom type
- sane defaults
- `apply` transformations
- ...

PR #233, opened 2021-04-09
core: add usbutils

blaggacao:
Looks like core does not include `lsusb` by default, now it does.

### DIFF
--- a/profiles/core/default.nix
+++ b/profiles/core/default.nix
@@ -28,6 +28,7 @@ in
       ripgrep
       skim
       tealdeer
+      usbutils
       utillinux
       whois
     ];


blaggacao:
I confirm: does as advised.

nrdxp (approved):
I'm not 100% on this, it's a pretty major archtectural shift, and I'm not sure it's in the right direction...
:laughing: 
(late april fools)
bors r+

blaggacao:
> I'm not 100% on this, it's a pretty major archtectural shift, and I'm not sure it's in the right direction...

With already two :wine_glass: intus, this is still even better of a kind of a review.

PR #234, opened 2021-04-09
evalFlakeArgs: fix userSpecialArgs defaults

Pacman99:
Just tried to drop the line and I realized the type for the attribute was wrong.


nrdxp:
You guys should be able to push directly to core for simple things like this. FYI

PR #236, opened 2021-04-10
init pkgs-lib and add tests and shell

Pacman99:
Towards the goal of removing top level folders and putting more things in lib. This shifts shell and tests to lib under the `pkgs-lib` namespace. This namespace is separated by architecture, because those lib functions rely on nixpkgs architecture specific derivations. I think this is the cleanest way to do it, you don't have to instantiate pkgs-lib for each arch, and it exposes some of these functions for others to use.

This PR also fixes multi-arch checks.


lib/pkgs-lib/shell/default.nix
```diff
@@ -1,25 +1,23 @@
...
-  }).${system};
+  overlays = [
+    inputs.devshell.overlay
+    (final: prev: {
+      deploy-rs =
+        inputs.deploy.packages.${prev.system}.deploy-rs;
+    })
+  ];
+
+  pkgs = dev.os.pkgImport nixos overlays system;
```
Pacman99, line 11: I thought I would just instantiate a nixpkgs for the specific tools here. Its not the greatest, but the alternative is to try and pass multiPkgs/pkgs to shell. Realistically this is the only place where these tools are needed. And I want to prevent passing pkgs to pkgs-lib to keep it more self contained.

Pacman99:
bors try

lib/pkgs-lib/tests/default.nix
```diff
@@ -1,17 +1,27 @@
...
+  mkChecks = hosts: nodes:
+    let
+      deployHosts = lib.filterAttrs
+        (n: _: hosts.${n}.config.nixpkgs.system == system)
+        nodes;
+      deployChecks = inputs.deploy.lib.${system}.deployChecks { nodes = deployHosts; };
+      tests = { libTests = libTests; }
+        // lib.optionalAttrs (deployHosts != { }) {
+        profilesTest = profilesTest (hosts.${(builtins.head (builtins.attrNames deployHosts))});
+      };
```
Pacman99, line 12: This allows the nixos and NixOS hosts/users to be dropped. It just looks for the first host which supports the architecture and runs the profilesTest with that. If profiles are written well enough, this should be fine.

Pacman99: Also this forces a deploy-rs dependency on lib, which I think is fine.

lib/mkFlake/default.nix
```diff
@@ -34,27 +34,21 @@ let
...
-            (n: _: self.nixosConfigurations.${n}.config.nixpkgs.system == system)
-            self.deploy.nodes;
-          deployChecks = deploy.lib.${system}.deployChecks { nodes = deployHosts; };
-        in
-        nixos.lib.recursiveUpdate tests deployChecks;
+      checks = pkgs-lib.tests.mkChecks {
+        inherit (self.deploy) nodes;
+        hosts = self.nixosConfigurations;
+        homes = self.homeConfigurations;
+      };
```
Pacman99, line 46: Just added home manager activation package checks

lib/pkgs-lib/shell/default.nix
nrdxp: I don't know enough about `nix` evaluation to know if this causes signficant overhead, or if it's smart enough to realize it already has the derivation on hand elsewhere. Either way, it's not too bad.

lib/mkFlake/default.nix
Pacman99: This will test all home-manager configurations that exist. I don't think this should be an issue.

lib/pkgs-lib/tests/default.nix
```diff
@@ -1,17 +1,29 @@
-{ self, pkgs }:
-let
-  inherit (self.inputs) nixos;
-  inherit (self.nixosConfigurations.NixOS.config.lib) testModule specialArgs;
+{ pkgs-lib, pkgs, system, inputs, nixos, lib, ... }:
+rec {
```
nrdxp, line 2: I try to avoid recursive attribute sets if at all possible as they can sometimes increase eval times and/or cause hard to debug infinite recursion errors. The one exception I've allowed so far was in suites, purely for convenience sake, and because factoring into a `let` binding there would be painful.

It looks like a `let` binding might work here instead though.

nrdxp (changes requested):

lib/pkgs-lib/tests/default.nix
Pacman99: That makes sense, updating with let/inherit

nrdxp (approved):
I think I reviewed everything thorough, so if something breaks you can blame me :laughing: 
bors r+

nrdxp (approved):
bors r+

blaggacao:
Nice! I like the inauguration of `pkgs-lib` concept.

I guess there are two mishaps inbhere, that can easily be fixed afterwards:

- If I remember correctly, `tests/lib.nix` would test `lib`, not `lib/pkgs-lib`. I guess it was moved to `lib/pkgs-lib/tests` because it depends on `pkgs`. However, the current place would generally indicate testing `pkgs-lib`.

- Similarily: `shell`. However, as far as I have been reasoning shell is not a library function of the devos library (_"export"_(, but a direct _"output"_ of the devos-lib.

PR #238, opened 2021-04-12
Implement evalArgs for new API

Pacman99:
This simply encodes the public facing portion #232. @blaggacao and I ended up changing the API after some discussion, so this is the most updated version of the planned API.


Pacman99:
closed to reserve api-next branch name for devos

PR #240, opened 2021-04-12
add mkFlakeDoc to pkgs-lib to build options doc

Pacman99:
Just built the doc - requires a bit of fen-angling - because current api-next flake doesn't work since mkflake hasn't been updated. We can use this PR to go over option/type descriptions and include any doc improvements. 

Overall I think the descriptions can use some work, we need to clearly communicate what each API argument is for and the reasoning behind the option. I think a fresh set of eyes could help with that, especially someone who wasn't part of all the discussion for the API design.

This document should be able to quickly answer questions like "What happened to extern?" or "How should I import modules from other flakes?". 

I should not attempt to cover everything - thats what the other doc sections are for - just a simple and straightforward summary of what each argument provides and is used for.

This replaces parts of #232.

### DIFF
--- a/doc/mkFlakeOptions.md
+++ b/doc/mkFlakeOptions.md
@@ -1,0 +1,323 @@
+## channels
+nixpkgs channels to create
+
+*_Type_*:
+attribute set of submodules
+
+*_Default_*
+```
+{"nixpkgs":{"input":"/nix/store/4dqrf3ly31g9ypgm89rn7b1p06vddll6-source"}}
+```
+
+
+## channels.<name>.config
+nixpkgs config for this channel
+
+*_Type_*:
+path that evaluates to a(n) attrs
+
+*_Default_*
+```
+{}
+```
+
+
+## channels.<name>.externalOverlays
+overlays to apply to the channel that don't get exported to the flake output
+useful to include overlays from inputs
+
+*_Type_*:
+path that evaluates to a(n) listOf
+
+*_Default_*
+```
+[]
+```
+
+
+## channels.<name>.input
+nixpkgs flake input to use for this channel
+
+*_Type_*:
+nix flake
+
+*_Default_*
+```
+"/nix/store/4dqrf3ly31g9ypgm89rn7b1p06vddll6-source"
+```
+
+
+## channels.<name>.overlays
+overlays to apply to this channel
+these will get exported under the 'overlays' flake output as <channel>/<name>
+
+*_Type_*:
+path that evaluates to a(n) listOf
+
+*_Default_*
+```
+[]
+```
+
+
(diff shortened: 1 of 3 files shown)


doc/mkFlakeOptions.md
```diff
@@ -0,0 +1,323 @@
...
+*_Type_*:
+attribute set of submodules
+
+*_Default_*
+```
+{"nixpkgs":{"input":"/nix/store/4dqrf3ly31g9ypgm89rn7b1p06vddll6-source"}}
+```
+
+
+## channels.<name>.config
```
Pacman99, line 13: The `<name>` part doesn't seem to get rendered right. I don't know enough about markdown to know why that is. Maybe I need to escape characters?

doc/mkFlakeOptions.md
```diff
@@ -0,0 +1,323 @@
...
+*_Type_*:
+path that evaluates to a(n) listOf
+
+*_Default_*
+```
+[]
+```
+
+
+## home
```
Pacman99, line 63: This is mentioned in evalArgs.nix but home doesn't get a `hosts` because homeConfigurations are auto-generated from nixos.hosts

doc/mkFlakeOptions.md
```diff
@@ -0,0 +1,323 @@
...
+
+*_Default_*
+```
+"<function>"
+```
+
+
+## self
+The flake to create the devos outputs for
+*_Type_*:
```
Pacman99, line 307: There should be a space between these two lines.

lib/pkgs-lib/default.nix
```diff
@@ -15,6 +15,7 @@ lib.genAttrs utils.lib.defaultSystems (system:
 
       tests = callLibs ./tests;
       shell = callLibs ./shell;
+      mkFlakeDoc = callLibs ./mkFlakeDoc.nix;
```
Pacman99, line 18: To build doc:
`nix build .#lib.pkgs-lib.x86_64-linux.mkFlakeDoc`

If we make lib a flake, this should go in it's `packages` output.

This doesn't build yet in the branch, you have to add the old evalArgs to get the flake to actually evaluate.

nrdxp:
This might need updating after release cleanup.

blaggacao:
I went ahead and added a rebased (but not optimized) version of this into the `api-next` branch [here](https://github.com/divnix/devos/pull/251/commits/24dbb2b3231b5a9f4a9aae57ef90436d59db300b).

It works as intended with `nix build .#jobs.mkFlakeDoc` (on `x86_64-linux`, only).

Hence, I'm going to close this so we can continue over there.

PR #245, opened 2021-04-18
create lib/flake.nix

blaggacao:
~~This commit doesn't actually activate lib checks as I don't know how to do that in a pre-flake world, the next PR #246 accomplishes this within the sub flake. (see question about how to hook up to CI over there).~~

Since #246 has been merged into this branch:

This PR factors out lib/tests and introduces lib/flake.nix


lib/tests/default.nix
```diff
@@ -0,0 +1,29 @@
...
+    export NIX_STATE_DIR=$TEST_ROOT/var/nix
+    export NIX_STORE_DIR=$TEST_ROOT/store
+    export PAGER=cat
+    cacheDir=$TEST_ROOT/binary-cache
+    nix-store --init
+
+    touch $out
+  '';
+in
+libTests
```
Pacman99, line 29: The `let`/`in` is unnecessary, you can just create the package directly, since libTests isn't recursive or being used by anything else.

lib/tests/lib.nix
```diff
@@ -0,0 +1,93 @@
+{ pkgs, lib, dev, ... }:
```
Pacman99, line 1: Not sure how this is supposed to be called, but if your not planning on using `callLib` the `...` is not needed.

Pacman99 (approved):
This looks great and logically makes a lot more sense, it was awkward to export tests in pkgs-lib when I first set that up.

lib/tests/lib.nix
blaggacao: It's even wrong, thanks!

blaggacao: Fixed in https://github.com/divnix/devos/pull/245/commits/68587aadc6cbc226f4c20c597b705664ef6ac824

lib/tests/default.nix
blaggacao: Fixed in new version of https://github.com/divnix/devos/pull/245/commits/99a6565535b6f5ba28f42937df50de6cc8232f20

lib/flake.nix
```diff
@@ -0,0 +1,72 @@
...
+
+  outputs = inputs@{ self, nixpkgs, deploy, devshell, flake-utils, ... }: let
+
+    inherit (nixpkgs) lib;
+
+    # We work with a combined lib, internally ...
+    combinedLib = lib.extend (final: prev:
+      let callLibs = file: import file
+        ({
+          lib = final;
```
Pacman99, line 23:
```suggestion
          lib = lib // { dev = final; }
```
And use lib.makeExtensible. I think thats clearer and helps towards removing callLibs. Also so we don't have to manually export things. And I think we should export all of devos lib, and only manually create the top level functions. Just don't want nixpkgs lib to be exported.

Pacman99: Although I guess this still keeps the lib vs dev issue. Then maybe we should just do `lib = recursiveUpdate lib final`

blaggacao:
I'm going ahead on this, since the lasting errors are surfacing previous issues that I'll not want to solve all in one single big bang PR. Issue:

```
nix flake check
error: anonymous function at /nix/store/fcdwyq3k0530vxi43fl6p9i8s71ibdb1-source/devos/mkPackages.nix:1:1 called without required argument 'userSelf'

       at /nix/store/fcdwyq3k0530vxi43fl6p9i8s71ibdb1-source/flake.nix:21:28:

           20|     combinedLib = lib.extend (final: prev:
           21|       let callLibs = file: import file
             |                            ^
           22|         ({
```

`userSelf` cannot be fed back from `mkFlake` into `callLib`.

PR #246, opened 2021-04-18
ref: extract lib into subflake

blaggacao:
Depends on #245 

In our current CI model, how would we actually hook the sub-flake checks up correctly? Will try to figure that out, now, but any quick mind solution welcome.


flake.nix
```diff
@@ -5,50 +5,50 @@
...
+      overlays = ./overlays;
+      profiles = ./profiles;
+      userProfiles = ./users/profiles;
+      modules = import ./modules/module-list.nix;
+      userModules = import ./users/modules/module-list.nix;
+    } // {
+    defaultTemplate = self.templates.flk;
+    templates.flk.path = ./.;
+    templates.flk.description = "flk template";
+  };
```
blaggacao, line 53: At this point, we don't want to export `lib` anymore in something that is to become a template.

flake.nix
```diff
@@ -5,50 +5,50 @@
     {
       nixos.url = "nixpkgs/nixos-unstable";
       override.url = "nixpkgs";
+      devos.url = "path:./lib"; # TODO: outfactor into separate repo
+      devos.inputs = {
+        nixpkgs.follows = "nixos";
```
blaggacao, line 10: Note: The library itself does not distinguish between nixos and nixpkgs. anything with decent library functions and trivial builders will do.

lib/devos/devosSystem.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, nixos, self, inputs, ... }:
+{ lib, nixpkgs, self, inputs, ... }:
```
blaggacao, line 1: Same here, this ought to come from function inputs rather than flake inputs.

lib/devos/mkHosts.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, dev, nixos, inputs, self, ... }:
+{ lib, dev, nixpkgs, inputs, self, ... }:
```
blaggacao, line 1: same here.

lib/devos/mkPkgs.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, dev, nixos, self, inputs, ... }:
+{ lib, dev, nixpkgs, self, inputs, ... }:
```
blaggacao, line 1: same here

lib/mkFlake/evalArgs.nix
```diff
@@ -54,10 +54,10 @@ let
         options = with types; {
           input = mkOption {
             type = flakeType;
-            default = inputs.nixos;
+            default = inputs.nixpkgs;
```
blaggacao, line 57: same here ff.

lib/pkgs-lib/default.nix
```diff
@@ -1,8 +1,8 @@
-args@{ lib, dev, utils, nixos, ... }:
+args@{ lib, dev, utils, nixpkgs, ... }:
```
blaggacao, line 1: not here, though.

lib/pkgs-lib/shell/default.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, dev, inputs, system, nixos, ... }:
+{ lib, dev, inputs, system, nixpkgs, ... }:
```
blaggacao, line 1: neither here

lib/pkgs-lib/tests/default.nix
```diff
@@ -1,4 +1,4 @@
-{ pkgs, system, inputs, nixos, lib, ... }:
+{ pkgs, system, inputs, nixpkgs, lib, ... }:
```
blaggacao, line 1: neither here

flake.nix
```diff
@@ -5,50 +5,53 @@
...
+      userProfiles = ./users/profiles;
+      modules = import ./modules/module-list.nix;
+      userModules = import ./users/modules/module-list.nix;
+    } // {
+    defaultTemplate = self.templates.flk;
+    templates.flk.path = ./.;
+    templates.flk.description = "flk template";
+
+    # until ./lib has own ci
+    checks.x86_64-linux.devos-lib-tests = devos.checks.x86_64-linux.tests;
```
blaggacao, line 55: I have no idea how to avoid that the whole `checks` will be overridden.

Maybe for the sake of `api-next` we can ignore normal system checks until we adopt the `fup`'s `anyMerge` capabilities in this context.

Pacman99: Why not just keep the lib tests in lib only? Why export it in devos's flake.nix. Only people hacking on lib care about those tests.

Also we can fix this problem if we utilize fup's otherArguments merging. We should be able to just pass `checks` to mkFlake as an argument and it will be exported right.

lib/flake.nix
```diff
@@ -0,0 +1,66 @@
...
+      test = import ./tests {
+        inherit (nixpkgs) pkgs;
+        inherit (self) lib;
+      };
+    };
+
+
+  in {
+
+    lib = lib.extend (final: prev:
```
Pacman99, line 31: If we use `lib.extend` then devos lib is going to be exporting all of nixpkgs lib along with it. Were you intending that?
I think makeExtensible would be the better option, like its done in lib/default.nix right now.

blaggacao: > Were you intending that?

Yes, the idea is to provide an onion style `lib` name space. It avoids the user to handle different libs for different purposes.

flake.nix
blaggacao: Yep, I'll take this out again. Since Tim is going to migrate to hercules, we can wire this up for our own purposes differently. The motivation was to have checks until we completely separate the repo.

lib/flake.nix
```diff
@@ -0,0 +1,66 @@
...
+
+        pkgs-lib = callLibs ./pkgs-lib;
+
+        inherit (attrs) mapFilterAttrs genAttrs' safeReadDir
+          pathsToImportedAttrs concatAttrs filterPackages;
+        inherit (lists) pathsIn;
+        inherit (strings) rgxToString;
+      });
+
+    checks.x86_64-linux.tests = jobs.test;
```
Pacman99, line 63: Can you try to make this multi-arch, we made sure lib tests doesn't have IFD so it should work fine.
Is `jobs` for hydra? Can that be made multi-arch?

blaggacao: > Is jobs for hydra?

it's proto-hydra, and shall be dispatched to `checks` and `hydraJobs` and other things.

> Can that be made multi-arch?

I'm not sure about `hydraJobs`.

> Can you try to make this multi-arch, we made sure lib tests doesn't have IFD so it should work fine.

Ok, I'll try. What's the motivation here? To test things like `flk` command on multiple arches and stuff like that?

blaggacao:
This has been merged into #245 . Still, let's keep the discussion here, I'll still update this branch accordingly.

PR #250, opened 2021-04-19
ref: reduce exposure to callLibs

blaggacao:
for clarity's sake, expose which function uses final and prev, so that
people can have a clearer understanding how they relate to each other
in terms of dependencies.

also a simple `{ lib = final; }` probably does not warrant a complete
callLibs obscurization.

### DIFF
--- a/lib/devos/mkPkgs.nix
+++ b/lib/devos/mkPkgs.nix
@@ -1,6 +1,8 @@
 { lib, utils }:
 
-{ extern, overrides, userFlakeNixOS, userFlakeSelf, userFlakeInputs }:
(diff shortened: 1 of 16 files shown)


lib/devos/mkPkgs.nix
```diff
@@ -20,6 +22,6 @@
       ++ extern.overlays
       ++ (lib.attrValues userFlakeSelf.overlays);
     in
-    { pkgs = lib.os.pkgImport nixpkgs overlays system; }
+    { pkgs = lib.os.pkgImport userFlakeNixOS overlays system; }
```
blaggacao, line 25: `userFlakeNixOS` here is correct: it represents the base line nixpkgs channel of the user.

PR #251, opened 2021-04-19
Intermediate version of api-next

blaggacao:
This is a pretty long series of internal refactorings in preparation of the stable API change, but that don't just yet change any public API.

I'd like to log this into `core` for the sake of reducing the (chore) commit inventory.


blaggacao:
bors try

blaggacao:
bors retry

blaggacao:
bors try -

blaggacao:
bors try-

blaggacao:
bors try

lib/flake.nix
```diff
@@ -0,0 +1,90 @@
...
+
+  in
+
+  {
+    inherit jobs;
+
+    lib = utils.lib // {
+      inherit (lib)
+        mkFlake;
+    };
```
Pacman99, line 70:
```suggestion
  inherit lib;
```
Is there a reason not to export the entire devos lib right now? Or even just move the lib down here and refer to it as `self.lib` when passing to checks.

lib/flake.nix
```diff
@@ -0,0 +1,90 @@
...
+
+    lib = nixpkgs.lib.makeExtensible (self:
+      let
+        attrs = import ./attrs.nix { lib = nixpkgs.lib // self; };
+        lists = import ./lists.nix { lib = nixpkgs.lib // self; };
+        strings = import ./strings.nix { lib = nixpkgs.lib // self; };
+      in
+
+      utils.lib
+
```
Pacman99, line 25: I like this! :+1: 
Could we recreate `callLibs` to just do this for us? `callLibs = file: import file { lib = nixpkgs.lib // self; }` So it has the same meaning as upstream lib.

lib/devos/mkHomeConfigurations.nix
```diff
@@ -1,12 +1,14 @@
-{ lib, self, ... }:
+{ lib }:
+
+{ userFlakeSelf }:
```
Pacman99, line 3: This should only take nixosConfigurations and as a regular input: `nixosConfgurations:`. That way it can be used by others too: `mkHomeConfigurations self.nixosConfigurations`.

lib/flake.nix
blaggacao: Nope, not really. I think before we finalize we should go over this again and only export stuff for which we identify a clear use case.

This, in oder to only commit (resources to maintain) a public api where we create actual value.

lib/flake.nix
blaggacao: We could not call it callLib though, since that would lead people to put more stuff into it (by analogy of upstream).

I have been thinking of `importWithNixpkgsLib` which would be appropriatly scoped, but then it's very likely easier to just grasp the meaning of `//`.

lib/devos/mkHomeConfigurations.nix
blaggacao: Correct! There are a couple of those still around which I didn't clean up just yet.

I was thinking we can address those while rebasing on `fup`.

blaggacao:
bors try

PR #254, opened 2021-04-22
rebase on fup and switch that API

blaggacao:


flake.nix
```diff
@@ -17,38 +17,81 @@
...
+          ];
+        };
+        latest = { };
+      };
+
+      os = {
+        hostDefaults = {
+          system = "x86_64-linux";
+          channelName = "nixos";
+          modules = ./modules;
```
Pacman99, line 61:
```suggestion
          modules = ./modules/module-list.nix;
```
this needs to be a list of modules, modules aren't auto imported. And for the same reason, you can't delete module-list.nix.

blaggacao: Ah true, this still represents my state of mind some two days ago :wink:

PR #258, opened 2021-04-22
Use GitHub actions for BORS Checks

nrdxp:


nrdxp:
bors try

Pacman99 (approved):
LGTM

.github/workflows/check.yml
```diff
@@ -0,0 +1,28 @@
+name: "Check & Cachix"
+on:
+  push:
+    branches:
+      - core
+      - trying
+      - staging
```
Pacman99, line 7: Can you add checks to develop? If more PR's target develop, I think this could be useful to run checks there too.

Pacman99: I take this back, it makes more sense to not run comprehensive checks on develop and just run them on the final PR from develop to core.

But it might still be good to include syntax checking on formatting check on develop.

Pacman99: although, merging PR's into develop with bors would be nice to follow the same merging strategy across the repo

blaggacao: In order to spare on github limits, shift-left strategy might be better, for time being.

blaggacao (approved):

.github/workflows/check.yml
nrdxp: I feel like github should have a way to persist certain volumes between runs but I haven't had time to figure it out just yet. If we could do that, we could target pretty much everything, even PR's and it wouldn't really be all that expensive.

blaggacao: https://levelup.gitconnected.com/github-actions-how-to-share-data-between-jobs-fc1547defc3e

PR #260, opened 2021-04-24
lib: one line for arguments, drop userFlake*

Pacman99:
Also format all files and add a flake.lock for lib

for a folder thats meant to work on other flakes theres never a reason it should need to refer to itself, only other flakes. So "self" and "inputs" are better namings for these variables. The userFlake* is redundant and confusing, when trying to call the functions its hard to figure out how to use them when there are two lines of arguments to figure out. 

Many of the functions don't even need this much information(ie mkHomeConfigurations and mkPkgs).

lib needs a flake.lock, because it is a flake we want others to  use. Also having to regenerate it every time its used makes no sense.

This also gets develop to pass `flake check` which it didn't because of deploy having a `utils.follows = "utils"`.

### DIFF
--- a/lib/jobs/mkFlakeDoc.nix
+++ b/lib/jobs/mkFlakeDoc.nix
@@ -22,9 +22,7 @@ let
   '';
 
   options = (
-    lib.mkFlake.evalArgs
-      { userFlakeSelf = {}; userFlakeNixOS = {}; }
-      { args = { }; }
+    lib.mkFlake.evalArgs { nixos = "nixos"; args = { }; }
   ).options;
 
   processedOptions = (pkgs.nixosOptionsDoc { inherit options; }).optionsNix;
--- a/flake.lock
+++ b/flake.lock
@@ -81,7 +81,7 @@
         "utils": "utils_2"
       },
       "locked": {
-        "narHash": "sha256-aGpcPxOBIPnwj6uLNtpKefZjEbDkZnMHZ2mH12veHnY=",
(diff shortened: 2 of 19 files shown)


blaggacao:
> This also gets develop to pass flake check which it didn't because of deploy having a utils.follows = "utils".

The reason of this is to avoid having two versions of utils, but since utils is really small, that's probably more a theoretical than a practical problem.

lib/jobs/mkFlakeDoc.nix
```diff
@@ -23,7 +23,7 @@ let
 
   options = (
     lib.mkFlake.evalArgs
-      { userFlakeSelf = {}; userFlakeNixOS = {}; }
+      { userFlakeSelf = { }; userFlakeNixOS = { }; }
```
blaggacao, line 26: This seems like an oversight. Is it?

blaggacao (approved):
LGTM, there _appears_ to be small oversight, if that makes sense.

lib/jobs/mkFlakeDoc.nix
Pacman99: yup oversight and :+1: on wording that as "appears", I'm going to use that.

lib/jobs/mkFlakeDoc.nix
```diff
@@ -22,9 +22,7 @@ let
   '';
 
   options = (
-    lib.mkFlake.evalArgs
-      { userFlakeSelf = {}; userFlakeNixOS = {}; }
-      { args = { }; }
+    lib.mkFlake.evalArgs { nixos = "nixos"; args = { }; }
```
Pacman99, line 25: This is just for docs, so "nixos" is fine.

PR #261, opened 2021-04-24
profilesTest/mkHosts: include host's configuration

Pacman99:
Now that profiles test is done dynamically - first available host - we likely need to include the host configuration itself. I have profiles that depend on the domain being set, which how I ran into this, and I think other problems might occur if we don't include the full host in the profiles test.

targeting core, since its a small bug fix and was planned to make into a release.

### DIFF
--- a/lib/pkgs-lib/tests/default.nix
+++ b/lib/pkgs-lib/tests/default.nix
@@ -48,7 +48,7 @@ let
     };
 
     testScript = ''
-      machine.systemctl("is-system-running --wait")
+      ${host.config.networking.hostName}.systemctl("is-system-running --wait")
     '';
   };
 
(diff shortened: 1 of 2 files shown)


lib/pkgs-lib/tests/default.nix
```diff
@@ -48,7 +48,7 @@ let
     };
 
     testScript = ''
-      machine.systemctl("is-system-running --wait")
+      ${host.config.networking.hostName}.systemctl("is-system-running --wait")
```
Pacman99, line 51: i don't know why `make-test` does this, apparently if a hostname is set in a node or machine, it uses that instead for the test script.. This is not documented either, and I wish it was just standard with the name of the node. But this is my workaround.

blaggacao (approved):

PR #262, opened 2021-04-24
move mkFlakeDoc to packages output also drop jobs

Pacman99:
I believe mkFlakeDoc belongs in pkgs-lib, because it is a lib related package/pkg. And with mkFlakeDoc in pkgs-lib there is no need for jobs. And I think anything that could go in jobs really should go in checks or packages. If something needs to be tested - checks, if something needs to be built - packages. jobs is not multi-arch and is redundant to build/test things with when official flake outputs exist. pkgs-lib/tests can house all the related tests and builds.


lib/pkgs-lib/default.nix
```diff
@@ -9,5 +9,10 @@ lib.genAttrs
     {
       tests = import ./tests { inherit lib deploy nixpkgs pkgs system; };
       shell = import ./shell { inherit lib devshell deploy nixpkgs system; };
+      mkFlakeDoc = pkgs.writeText "mkFlakeOptions.md" (
+        pkgs.nixosOptionsDoc {
+          inherit (lib.mkFlake.evalArgs { nixos = "nixos"; args = { }; }) options;
+        }
+      ).optionsMDDoc;
```
Pacman99, line 16: added `optionsMDDoc` upstream, so this is much smaller, and doesn't really deserve its own file.

PR #263, opened 2021-04-24
create lib.modules to store all devos modules

Pacman99:
Currently devosSystem and mkHosts a few modules in themselves. This ends up with them containing lots of code thats not directly related to its purpose. So I extracted those modules into their own file `lib/modules.nix` and namespace `lib.modules`. Each attribute is a module or a function to a module to describe what each module depends on. These are then used in devosSystem and mkHosts.

This should make rebase on fup easier, you can just drop lines in each module or remove unnecessary dependencies much easier if they have their own dedicated file.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -81,7 +81,7 @@
         "utils": "utils_2"
       },
       "locked": {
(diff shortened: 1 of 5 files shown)


blaggacao (approved):
Nice! Good chore work!

PR #264, opened 2021-04-26
Implement new api

Pacman99:
This is the underlying implementation of the api thats defined in the new `evalArgs`, it drops and simplifies a lot of the code in devos.
Obsoleted by fup:
 - mkHosts => you can just pass modules, specialArgs, and builder to hostDefaults instead
 - mkPkgs => fup creates the multiPkgs and can handle applying overlays and also has the benefit of multiple channels
 - mkPackages => There is a better implementation in fup and we can now use `packagesBuilder`.

This includes many other changes to get to the final rebase - explanation is in commit messages.


lib/devos/devosSystem.nix
```diff
@@ -1,31 +1,33 @@
...
-let args = builtins.removeAttrs allArgs [ "self" "nixos" "inputs" ]; in
-lib.nixosSystem (args // {
-  modules =
-    let
-      moduleList = builtins.attrValues modules;
+# dependencies to return a builder
+{ self, inputs }:
 
-      fullHostConfig = (lib.nixosSystem (args // { modules = moduleList; })).config;
+{ modules, specialArgs, ... } @ args:
```
blaggacao, line 6: `specialArgs` &rarr; `channel`?

lib/devos/devosSystem.nix
```diff
@@ -1,31 +1,33 @@
...
+specialArgs.channel.input.lib.nixosSystem
+  (args // {
+    modules =
+      let
+        fullHostConfig = (lib.nixosSystem (args // { inherit modules; })).config;
```
blaggacao, line 11: `lib = channel.input.lib;` is that missing?

blaggacao (approved):
Two quick remarks, otherwise a massive improvement, we should merge roght away.

lib/devos/devosSystem.nix
Pacman99: the plan is for channel to be in specialArgs as per gytis-ivaskevicius/flake-utils-plus#44

lib/devos/devosSystem.nix
Pacman99: yes oversight, I'll probably do a let at the top to set `nixosSystem`

lib/devos/devosSystem.nix
blaggacao: oh, got it on second revise

flake.nix
```diff
@@ -17,38 +17,74 @@
...
+      channelsConfig = { allowUnfree = true; };
+
+      channels = {
+        nixos = {
+          overlays = [
+            (devos.lib.pathsIn ./overlays)
+            ./pkgs/default.nix
+            pkgs.overlay # for `srcs`
+            nur.overlay
+          ];
```
Pacman99, line 49: lists get flattened for overlays and modules, but not `externalModules` and `hostDefaults.modules`.

lib/mkFlake/default.nix
```diff
@@ -1,64 +1,94 @@
...
 
-    nixosModules = cfg.modules;
+  # evalArgs sets channelName and system to null by default
+  # but for proper default handling in fup, null args have to be removed
+  stripHost = args: removeAttrs (lib.filterAttrs (_: arg: arg != null) args) [
+    # arguments in our hosts/hostDefaults api that shouldn't be passed to fup
+    "externalModules"
+  ];
+  hosts = lib.mapAttrs (_: stripHost) cfg.nixos.hosts;
+  hostDefaults = stripHost cfg.nixos.hostDefaults;
```
Pacman99, line 41: this, however, is necessary since host arguments are validated. And to let fup handle defaults, we have to remove anything thats null - default value for channelName and system.

overlays/overrides.nix
```diff
@@ -0,0 +1,26 @@
+channels: final: prev: {
```
Pacman99, line 1: this file is simply an overlay now and nothing special, so it makes sense to include in the overlays folder.

Pacman99 (commented):
Highlighted some important parts of these changes. But overall I think its ready for final review and merge. I don't have anything else to include in here.

flake.nix
```diff
@@ -17,38 +17,74 @@
...
+          externalModules = [
+            ci-agent.nixosModules.agent-profile
+            home.nixosModules.home-manager
+          ];
+        };
+        hosts = {
+          NixOS = {
+            modules = ./hosts/NixOS.nix;
+          };
+        };
```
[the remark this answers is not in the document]
Pacman99: actually I could just include this in the PR, I think such a function would be very similar to `recImport`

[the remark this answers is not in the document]
Pacman99: added this in https://github.com/divnix/devos/pull/264/commits/aa825b87a6a16a8a663e14f52155bc924d28bad1

not going to do it in this PR, but a dedicated file for devos importers would be really nice

flake.nix
```diff
@@ -17,38 +17,73 @@
...
+          modules = ./modules/module-list.nix;
+          externalModules = [
+            ci-agent.nixosModules.agent-profile
+            home.nixosModules.home-manager
+          ];
+        };
+        hosts = nixos.lib.mkMerge [
+          (devos.lib.importHosts ./hosts)
+          { /* set host specific properties here */ }
+        ];
```
Pacman99, line 67: auto-import magic is clearly visible in the template, so its easy to override and even remove.

blaggacao: Nice, I'd silently wondered how to do this on my last review.

PR #266, opened 2021-04-28
pass builderArgs to config and extract builds

### LINKED ISSUE #170: Extract custom system builds from devosSystem out of lib
There should be some way for users to add and edit their own custom system builds for all hosts like what is currently being done for `isoConfig` and in the future `hmConfig`. 

Edit lib/devos/devosSystem.nix. But optimally the end user shouldn't have to edit a lib function.

## Additional context
I think this could be done as a file that defines an attrset mapping builds to nixosSystem's(or darwinSystems, homeConfigurations, robotnixConfigurations). And all builds will be available under `config.system.builds` as is currently being done. 

related: #168 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/97544759-extract-custom-system-builds-from-devossystem-out-of-lib?utm_campaign=plugin&utm_content=tracker%2F161955272&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F161955272&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Pacman99:
have mkHomeConfiguration create its own custom build within the function
create a externalModule for customBuilds so its easy to add more

closes #170

### DIFF
--- a/lib/mkFlake/default.nix
+++ b/lib/mkFlake/default.nix
@@ -64,7 +64,10 @@ lib.systemFlake (lib.mergeAny
     hostDefaults = lib.mergeAny hostDefaults {
       specialArgs.suites = cfg.nixos.suites;
       modules = cfg.nixos.hostDefaults.externalModules ++ defaultModules;
-      builder = os.devosSystem { inherit self inputs; };
+      builder = args: args.specialArgs.channel.input.lib.nixosSystem (lib.mergeAny args {
+        # So modules and functions can create their own version of the build
+        modules = [ { lib.builderArgs = args; } ];
+      });
     };
 
     nixosModules = lib.exporter.modulesFromList cfg.nixos.hostDefaults.modules;
--- a/modules/customBuilds.nix
+++ b/modules/customBuilds.nix
@@ -1,0 +1,29 @@
+{ lib, self, devlib, config, modules, channel, ... }:
+let
+  mkBuild = buildModule:
+    channel.input.lib.nixosSystem (devlib.mergeAny config.lib.builderArgs {
+      modules = [ buildModule ];
+    });
+in
+{
+  system.build = {
+    iso = (mkBuild (devlib.modules.isoConfig {
+      inherit self;
+      inherit (self) inputs;
+      fullHostConfig = config;
+    })).config.system.build.isoImage;
+
+    homes = (mkBuild ({ config, ... }: {
+      home-manager.useUserPackages = lib.mkForce false;
+      home-manager.sharedModules = [
+        {
+          home.sessionVariables = {
+            inherit (config.environment.sessionVariables) NIX_PATH;
+          };
+          xdg.configFile."nix/registry.json".text =
+            config.environment.etc."nix/registry.json".text;
+        }
+      ];
+    })).config.home-manager.users;
+  };
+}
--- a/flake.lock
+++ b/flake.lock
@@ -79,7 +79,7 @@
         "utils": "utils_2"
       },
       "locked": {
-        "narHash": "sha256-TQvd6TvSuT0sJCLlGsV65YjB+nIfDdDKZ1F94pCfkTw=",
+        "narHash": "sha256-oTiKYoR210VwjomzfSn/pCJ3immdUDRUPbYTDGaPFn8=",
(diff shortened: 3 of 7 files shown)


lib/devos/devosSystem.nix
```diff
@@ -1,40 +1,13 @@
...
-        system.build = {
-          iso = isoConfig.system.build.isoImage;
-          homes = hmConfig.home-manager.users;
-        };
         lib = {
-          inherit specialArgs;
-          testModule = {
-            imports = modules;
-          };
+          builderArgs = args;
```
Pacman99, line 10: This is now the only thing `devosSystem` does, I'm not sure there needs to be a separate file for it.

lib/mkFlake/default.nix
```diff
@@ -23,6 +23,7 @@ let
     (globalDefaults {
       inherit self inputs;
     })
+    { _module.args.devlib = lib; }
```
blaggacao, line 26: Why not `lib` as we use it in other files?

Pacman99: I tried that and it results in some weird issues. I think its a bit dangerous to override `lib` since every module uses it and our "lib" might not have merged with nixpkgs lib properly.

modules/customBuilds.nix
```diff
@@ -0,0 +1,10 @@
+{ self, devlib, config, modules, channel, ... }:
+{
+  system.build = {
+    iso = (devlib.os.mkBuild config (devlib.modules.isoConfig {
```
blaggacao, line 4: `lib.modules.iso` ? &mdash; that was the token deleted in the other file.

lib/mkFlake/default.nix
Pacman99: Even if we can get it to merge right, its still dangerous imo to override such a core component of nixos. If we even make a small mistake in merging or our lib it could break all hosts.

modules/customBuilds.nix
Pacman99: `isoConfig` was deleted in `devosSystem` not lib.modules. The `lib.modules.isoConfig` function still exists.

lib/devos/default.nix
```diff
@@ -7,6 +7,15 @@
       config = { allowUnfree = true; };
     };
 
+  mkBuild = config: module:
```
blaggacao, line 10:
```suggestion
  mkBuild = config: buildModule:
```

Just to make it a tiny bit more obvious tht we are talking about a specific type of module, not "any" module.

modules/customBuilds.nix
blaggacao: Maybe there was an issue with the previous code, then: https://github.com/divnix/devos/pull/266/files#diff-82b93923ced769e1c083b4982e8e68c75b5bca3d34620c2fcf22efe3f6cb4307L17

Cool! It looks better with `isoConfig`.

lib/devos/default.nix
Pacman99: done in aa17170

lib/mkFlake/default.nix
blaggacao: Yep, I agree, this is public facing, I forgot.

modules/customBuilds.nix
Pacman99: Ohh yeah I guess isoConfig has been broken for a bit.

lib/devos/devosSystem.nix
blaggacao: I'd say, no need for a separate file, look like a lot of teasing for little content. :wink:

Pacman99: done 45073c6

Pacman99:
Updates:
 - Moved mkBuild and `hmConfig` to be in customBuilds. So theres now a dedicated location for everything "build" related. I don't see a reason to use mkBuild outside of `customBuilds` and it could even be harmful to do so.
 - set _module.args.devlib in `globalDefaults`
 - made mkHomeConfigurations more portable, so it can be used for non-devos hosts

PR #268, opened 2021-04-29
use __dontExport property for lib and overrides

Pacman99:


blaggacao (approved):
LGTM! Excellent solution!

PR #271, opened 2021-05-02
Rearrange and cleanup lib

Pacman99:
Improve the flake.nix of lib, export some more functions, and remove obsoleted code


blaggacao (approved):
Nice! Is this already the final touch?

I suggest to make sure to also run `nixpkgs-fmt`, in case you not already have.

Pacman99:
> Nice! Is this already the final touch?

I believe so, I think were close to the _current_ finish line. I have one more change to move homeConfigurations and deploy.nodes to be set in the template's flake.nix instead of in `mkFlake` with a bit of improvements for both functions.

> I suggest to make sure to also run nixpkgs-fmt, in case you not already have.

done

lib/flake.nix
```diff
@@ -10,60 +10,49 @@
...
     {
-      lib = utils.lib // {
+      lib = with lib; utils.lib // {
+        inherit attrs lists modules importers generators;
         inherit (lib)
-          mkFlake pathsIn importHosts;
+          mkFlake
+          pathsIn
+          importHosts
+          mkDeployNodes;
```
Pacman99, line 54: Thoughts on public exports? I opted to export everything under their namespaces and just those four functions in top-level. I'll probably add mkHomeConfigurations too since I'm considering to export `homeConfigurations` in the template's flake.nix

blaggacao: I think the general direction here for export is already decent and considerate. I'm very glad we have arrived at this stance. :heart:

It makes sense to add these.

blaggacao:
Would you want to merge once you feel it's ready? (Please do)

PR #272, opened 2021-05-02
improve home-manager and deploy-rs integration

Pacman99:
move both to be setup in template
only set home-manager options if they exist

### DIFF
--- a/lib/generators.nix
+++ b/lib/generators.nix
@@ -1,4 +1,4 @@
-{ lib }:
+{ lib, deploy }:
 {
   mkHomeConfigurations = nixosConfigurations:
     with lib;
@@ -12,7 +12,7 @@
     in
     foldl recursiveUpdate { } (attrValues hmConfigs);
 
-  mkDeployNodes =
+  mkDeployNodes = hosts: extraConfig:
     /**
       Synopsis: mkNodes _nixosConfigurations_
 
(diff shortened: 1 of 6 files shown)


lib/generators.nix
```diff
@@ -12,22 +12,22 @@
...
+  mkDeployNodes = hosts: extraConfig:
     /**
       Synopsis: mkNodes _nixosConfigurations_
 
       Generate the `nodes` attribute expected by deploy-rs
       where _nixosConfigurations_ are `nodes`.
       **/
 
-    deploy: lib.mapAttrs (_: config: {
+    lib.mapAttrs (_: config: lib.recursiveUpdate extraConfig {
```
Pacman99, line 23: I want to be able to customize deploy-rs settings easily and with the move for integrations to be exposed in the template, I think this is the best way to do that.

lib/generators.nix
```diff
@@ -12,22 +12,22 @@
     in
     foldl recursiveUpdate { } (attrValues hmConfigs);
 
-  mkDeployNodes =
+  mkDeployNodes = hosts: extraConfig:
```
Pacman99, line 15: I want to be able to customize deploy-rs settings easily and with the move for integrations to be exposed in the template, I think this is the best way to allow that.

blaggacao (approved):
Good cleanup!

PR #273, opened 2021-05-05
Switch to out of tree devlib

Pacman99:
Devos lib has been extracted in https://github.com/divnix/devlib, so this PR switches to use that.

To still allow users to have their own lib, I added the `ourlib` concept that gets added in `sharedOverlays` and as a module argument.

I was originally planning on just reusing `devlib` , but I think I like this better.


blaggacao (approved):
LGTM. Thank you!

PR #275, opened 2021-05-06
shell: override nix to unstable for nixos-rebuild

### LINKED ISSUE #274: How to get Started?
Hi!
I'm a Nix and NixOS novice but I'm very intrigued by devos and the idea of "standard issue" operating systems and dev environments (especially if they could be automated as WSL VM's). But I'm having a bit of trouble getting started on my (bare metal) NixOS 20.09 installation.

I assumed that if I follow the instructions at https://devos.divnix.com/start/index.html that... well, I'm not sure what will happen actually, but I was hoping it would be obvious when I got there :sweat_smile: 

But instead I get the following:
```sh
$ nix-shell https://github.com/divnix/devos/archive/core.tar.gz -A shell --run "flk get core"
unpacking 'https://github.com/divnix/devos/archive/core.tar.gz'...
error: opening file '/nix/store/0b66qq08r921pbr40qrajidqqsknh3gs-xscreensaver.pam.lock': Permission denied
```

When I run the same command with `sudo` I apparently run out of memory
```
sudo nix-shell https://github.com/divnix/devos/archive/core.tar.gz -A shell --run "flk get core"
[sudo] password for d4hines: 
warning: dumping very large path (> 256 MiB); this may run out of memory
error: out of memory
```
(this machine has 64GB of memory :thinking:)

Then I had a big-brain moment: this is Nix, and there's a `default.nix` file in the root, so I should be able to just clone it and build it, ya? But alas, I get different errors depending on what I try. From my default nix-shell, I cloned the repo, and ran `nix-shell` and I was greeted with `🔨 Welcome to devshell`, which was a good sign!

`flk up` worked. However, `flk nixos switch` gave the the following error output:
```
[sudo] password for d4hines: 
error: unrecognised flag '--experimental-features'
Try 'nix --help' for more information.
```

That's _very_ discouraging, as the whole point of trying out Nix was to avoid issues like this. Nonetheless, I will keep trying! On a hunch, I tried removing all references to `experimental-features` in the repo, but this had no effect.

Any help would be greatly appreciated!

Pacman99:
fixes #274

### DIFF
--- a/lib/pkgs-lib/shell/default.nix
+++ b/lib/pkgs-lib/shell/default.nix
@@ -6,6 +6,11 @@ let
       deploy-rs =
         inputs.deploy.packages.${prev.system}.deploy-rs;
     })
+    (final: prev: {
+      nixos-rebuild = prev.nixos-rebuild.override {
+        nix = prev.nixFlakes;
+      };
+    })
   ];
 
   pkgs = dev.os.pkgImport nixos overlays system;
@@ -24,6 +29,7 @@ pkgs.devshell.mkShell {
     nixos-install
     nixos-generate-config
     nixos-enter
+    pkgs.nixos-rebuild
   ];
 
   git.hooks = {


nrdxp:
in order to test this properly, would I have be using stable nix?

Pacman99:
Yeah, nixos-rebuild by default in nixos gets installed with it's `nix` dependency being overridden to `config.nix.package`. So this problem would only occur if you haven't yet set `nix.package` to `pkgs.nixFlakes`.

Pacman99:
Converting to draft, I would prefer just getting #276 in.

Pacman99:
Fixed in core

PR #279, opened 2021-05-13
Agenix integration

Pacman99:
AFAICT This is mostly a documentation thing. But with divnix/devlib#2 we can now add agenix to the devshell.

### DIFF
--- a/doc/secrets.md
+++ b/doc/secrets.md
@@ -1,18 +1,110 @@
 # Secrets
-Secrets are managed using [git-crypt][git-crypt] so you can keep your flake in
-a public repository like GitHub without exposing your password or other
-sensitive data.
+Secrets are managed using [git-crypt][git-crypt] and [agenix][agenix]
+so you can keep your flake in a public repository like GitHub without
+exposing your password or other sensitive data.
 
 By default, everything in the secrets folder is automatically encrypted. Just
 be sure to run `git-crypt init` before putting anything in here.
 
+## Agenix
+Currently, there is [no mechanism][secrets-issue] in nix itself to deploy secrets
+within the nix store because it is world-readable.
+
+Most NixOS modules have the ability to set options to files in the system, outside
+the nix store, that contain sensitive information. You can use [agenix][agenix]
+to easily setup those secret files declaratively.
+
+[agenix][agenix] encrypts secrets and stores them as .age files in your repository.
+Age files are encrypted with multiple ssh public keys, so any host or user with a
+matching ssh private key can read the data. The [age module][age module] will add those
+encrypted files to the nix store and decrypt them on activation to `/run/secrets`.
+
+### Setup
+All hosts must have openssh enabled, this is done by default in the core profile.
+
+You need to populate your `secrets/secrets.nix` with the proper ssh public keys.
+Be extra careful to make sure you only add public keys, you should never share a
+private key!!
+
+secrets/secrets.nix:
+```nix
+let
+  system = "<system ssh key>";
+  user = "<user ssh key>";
+  allKeys = [ system user ];
+in
+```
+
+On most systems, you can get your systems ssh public key from `/etc/ssh/ssh_host_ed25519_key.pub`. If
+this file doesn't exist you likely need to enable openssh and rebuild your system.
+
+Your users ssh public key is probably stored in `~/.ssh/id_ed25519.pub` or
+`~/.ssh/id_rsa.pub`. If you haven't generated a ssh key yet, be sure do so:
+```sh
+ssh-keygen -t ed25519
+```
+
 > ##### _Note:_
-> Currently, there is [no mechanism][secrets-issue] in nix to deploy secrets
-> within the nix/store so, if they end up in the nix/store after deployment, they
-> will be world readable on that machine.
->
-> The author of devos intends to implement a workaround for this situation in
-> the near future, but for the time being, simple be aware of this.
+> The underlying tool used by agenix, rage, doesn't work well with password protected
+> ssh keys. So if you have lots of secrets you might have to type in your password many
+> times.
+
+
+### Secrets
+You will need the `agenix` command to create secrets. DevOS conveniently provides that
+in the devShell, so just run `nix develop` whenever you want to edit secrets. Make sure
+to always run `agenix` while in the `secrets/` folder, so it can pick up your `secrets.nix`.
+
+To create secrets, simply add lines to your `secrets/secrets.nix`:
+```
+let
+  ...
+  allKeys = [ system user ];
+in
+{
+  "secret.age".publicKeys = allKeys;
+}
+```
+That would tell agenix to create a `secret.age` file that is encrypted with the `system`
+and `user` ssh public key.
+
+Then go into the `secrets` folder and run:
+```sh
+agenix -e secret.age
+```
+This will create the `secret.age`, if it doesn't already exist, and allow you to edit it.
+
+If you ever change the `publicKeys` entry of any secret make sure to rekey the secrets:
+```sh
+agenix --rekey
+```
+
+### Usage
+Once you have your secret file encrypted and ready to use, you can utilize the [age module][age module]
+to ensure that your secrets end up in `/run/secrets`.
+
+In any profile that uses a NixOS module that requires a secret you can enable a particular secret like so:
+
+```nix
+{ self, ... }:
+{
+  age.secrets.mysecret.file = "${self}/secrets/mysecret.age";
+}
+```
+
+
+Then you can just pass the path `/run/secrets/mysecret` to the module.
+
+You can make use of the many options provided by the age module to customize where and how
+secrets get decrypted. You can learn about them by looking at the
+[age module][age module].
+
+
+> ##### _Note:_
+> You can take a look at the [agenix repository][agenix] for more information
+> about the tool.
 
 [git-crypt]: https://github.com/AGWA/git-crypt
+[agenix]: https://github.com/ryantm/agenix
+[age module]: https://github.com/ryantm/agenix/blob/master/modules/age.nix
 [secrets-issue]: https://github.com/NixOS/nix/issues/8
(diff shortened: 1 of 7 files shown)


doc/secrets.md
```diff
@@ -1,18 +1,121 @@
...
+
+First ensure all hosts have the openssh daemon enabled:
+
+profiles/core/default.nix:
+```nix
+{
+  services.openssh.enable = true;
+  services.openssh.openFirewall = false;
+}
+```
```
Pacman99, line 29: should we just do this by default?

blaggacao: Yes

secrets/.gitattributes
```diff
@@ -1,3 +1,4 @@
 * filter=git-crypt diff=git-crypt
 .gitattributes !filter !diff
+secrets.nix !filter !diff
 README.md !filter !diff
```
Pacman99, line 4: `git-crypt` is a little pointless if your using agenix properly. I also don't really like using it in general, agenix is just a much better solution for nix based setups. We should probably leave it in the template for backwards compatibility - users might be storing secrets with the assumption that they are going to be encrypted.

But I would be in favor of removing it at some point.

Cool addition if you drop git-crypt: `sudo nixos-rebuild switch --flake github:<nix config repo>`.

doc/secrets.md
```diff
@@ -1,18 +1,109 @@
...
+secrets/secrets.nix:
+```nix
+let
+  system = "<system ssh key>";
+  user = "<user ssh key>";
+  allKeys = [ system user ];
+in
+```
+
+You can get your systems ssh public key from `/etc/ssh/ssh_host_ed25519_key.pub`. If
```
blaggacao, line 35: Is this true for all systems with openssh?

Pacman99: Yeah I believe its the ssh daemon default. I can change the line to "you can most likely get your ...", but I'm not sure its necessary. If someone changed this directory, they will definitely remember.

doc/secrets.md
```diff
@@ -1,18 +1,109 @@
...
+this file doesn't exist you likely need to enable openssh and rebuild your system.
+
+Your users ssh public key is probably stored in `~/.ssh/id_rsa.pub`. If you haven't
+generated a ssh key yet, be sure do so:
+```sh
+ssh-keygen -t rsa
+```
```
blaggacao, line 42: We probably should encourage `ed25519` instead of `rsa` in the docs?
Just because so we can pretend to care. :wink: 

See: https://medium.com/risan/upgrade-your-ssh-key-to-ed25519-c6e8d60d3c54

Pacman99: Ohh yes definitely, fixed

secrets/.gitattributes
blaggacao: I'm also in favor of dropping it.

blaggacao (approved):
Super exciting!

doc/secrets.md
```diff
@@ -1,18 +1,109 @@
...
 > ##### _Note:_
-> Currently, there is [no mechanism][secrets-issue] in nix to deploy secrets
-> within the nix/store so, if they end up in the nix/store after deployment, they
-> will be world readable on that machine.
->
-> The author of devos intends to implement a workaround for this situation in
-> the near future, but for the time being, simple be aware of this.
+>  - Currently, there is [no mechanism][secrets-issue] in nix to deploy secrets
+>    within the nix/store so, if they end up in the nix/store after deployment, they
+>    will be world readable on that machine.
```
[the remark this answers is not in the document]
Pacman99: I decided to make that line the intro to the agenix section. Since that is the problem the document is explaining how to solve.

profiles/core/default.nix
```diff
@@ -143,6 +143,12 @@ in
     '';
   };
 
+  # For rage encryption, all hosts need a ssh key pair
```
GTrunSec, line 146: The practice of agenix should enable the `ed25519` in ssh services by default. like:
```
  services.openssh.hostKeys = [
    { type = "rsa"; bits = 4096; path = "/etc/ssh/ssh_host_rsa_key"; }
    { type = "ed25519"; bits = 256; path = "/etc/ssh/ssh_host_ed25519_key"; }
  ];
```

GTrunSec:
I did testing of secrets between agenix and sops-nix.

Maybe those commands will help users to use agenix or sops easily.
https://github.com/GTrunSec/nixops-infra-land/blob/b1fb9983f9ce6a58691e87410827876167fcfc35/devshell.toml#L16

Default setting of ssh:

- https://github.com/GTrunSec/nixops-infra-land/blob/b1fb9983f9ce6a58691e87410827876167fcfc35/profiles/core/core-services.nix#L4

what do you think of this suggestion?

profiles/core/default.nix
Pacman99: Is this necessary? Looking at `man configuration.nix` the default for `hostKeys` is just what you put except without the `bits = 256;` part.

Pacman99:
Thanks for the suggestions!

> https://github.com/GTrunSec/nixops-infra-land/blob/b1fb9983f9ce6a58691e87410827876167fcfc35/devshell.toml#L16

That looks useful, but for agenix I don't know if its warranted to add an entire command in the template to get a hosts public key. Its not something most people would do often. Although more pressing, this version of the template doesn't allow as much devshell customization. Something I would like to fix eventually.

profiles/core/default.nix
GTrunSec: Yes, Sorry I overlooked the man page. Just ignore this comment.

Pacman99: Would be nice if it was possible to generate these files without enabling the openssh daemon. It would require decoupling the key generation parts of the nixos sshd module.

Pacman99:
I think we can do this here too
bors try

Pacman99:
> https://github.com/GTrunSec/nixops-infra-land/blob/b1fb9983f9ce6a58691e87410827876167fcfc35/devshell.toml#L16

I'm thinking about this again. And I realize the command is not agenix specific and generally pretty useful. I think it would be a good addition to the shell and can be done in in the [library](https://github.com/divnix/digga) itself. Would you mind making a PR there to add it?

profiles/core/default.nix
GTrunSec: How about if I write a bash to generate these files by executing remote ssh. It would be an alternative choice for generating keys.

GTrunSec:
>I think it would be a good addition to the shell and can be done in in the library itself. Would you mind making a PR there to add it?

Sure, by the way, Do you have a plan to add the sops-nix module for extending secrets?

Pacman99:
> Sure, by the way, Do you have a plan to add the sops-nix module for extending secrets?

Not currently, I also don't entirely understand how to use sops-nix. I opted for agenix here due to simplicity. 

For the most part secrets management is documentation, so I'm sure a doc addition to explain sops-nix usage with devos would be very welcome. We could discuss adding sops-nix to the template by default with such a change.

blaggacao:
When comparing `sops-nix` and `agenix`, I cannot see how `sops-nix` could outperform `agenix` or even complement with additional features.

Maybe there is a complementary element which I havn't been able to recognize, yet.

As such, I would think we can relatively safely limit our attention to agenix. No?

Pacman99:
> Maybe there is a complementary element which I havn't been able to recognize, yet.

This is what I was thinking. And it looks like @GTrunSec's repo shows that. They used agenix and sops-nix together. At the very least I think documentation on how to use sops-nix with devos would be good. And I would be open to discussion on how it can be integrated in the template along with agenix. 

Also I know we are considering multiple templates in the future. So maybe one of those could have sops-nix.

profiles/core/default.nix
Pacman99: Not a bad idea, but it doesn't feel declarative. I think this seems to be the best solution for now. We could alternatively generate the hostKeys ourself, but that would conflict with the ssh module.

blaggacao:
Agree, while let's remember:

As long as we maintain clear, and well-argumented preferences, any documented alternative still is pareto-dominated, as it doesn't burden an unfeasible choice especially on our newcomer user persona (let's him "Novatus" from now on).

blaggacao:
I'm starting on putting on my "Novatus" cap, while reviewing a little our documentation.

What's holing this back? I think it is a very useful improvement.

Pacman99:
> What's holing this back? I think it is a very useful improvement.

I think its ready to go, feel free to merge. I think devshell additions and any other agenix features can be added in separate PR's.

> I'm putting on my "Novatus" cap, while I start to review our documentation.

I think more extensive doc changes should target develop. A little to prevent merge conflicts, but also to improve the review process of doc changes.

blaggacao (approved):

blaggacao:
Looking at it from https://github.com/divnix/digga/pull/20 angle, we probably wanted to implement some of the devshell enhanceents through https://github.com/divnix/digga/pull/11 for agenix. :stuck_out_tongue_winking_eye:

Pacman99:
@GTrunSec's comments were one of my main motivations for divnix/digga#11. I would be for adding add a devshell.toml to this repo.

profiles/core/default.nix
nrdxp: For future reference, perhaps we could just make a more minimal module that generates the keys, but doesn't enable ssh.

secrets/.gitattributes
nrdxp: I'm late to the party, but it would be good to drop as it actually enourages bad secrets management (putting secrets into Nix store). This was really more of a move of desparation on my part than an actual well thought out secrets strategy.

Pacman99: definitely agree, I'm just more worried about someone not realizing git-crypt usage in secrets is dropped and committing their secrets to a public repo. Which is arguably worse.

profiles/core/default.nix
Pacman99: Optimally we could contribute changes to the upstream ssh module to separate host key generation from the daemon itself.

PR #281, opened 2021-05-15
pkgs-lib/shell: sudo nixos-generate-config

bb010g:
```
$ btrfs subvol show / >/dev/null; printf "%s\n" "$?"
ERROR: Could not search B-tree: Operation not permitted
1
```

Btrfs's `ioctl()` `BTRFS_IOC_TREE_SEARCH` requires elevation. `btrfs subvol show <path>` may be called by `nixos-generate-config`, and if `btrfs subvol show <path>` fails then `nixos-generate-config` fails.

### DIFF
--- a/lib/pkgs-lib/shell/flk.sh
+++ b/lib/pkgs-lib/shell/flk.sh
@@ -33,7 +33,8 @@ case "$1" in
   "up")
     mkdir -p "$DEVSHELL_ROOT/up"
 
-    nixos-generate-config --dir "$DEVSHELL_ROOT/up/$HOSTNAME"
+    # `sudo` is necessary for `btrfs subvolume show`
+    sudo nixos-generate-config --dir "$DEVSHELL_ROOT/up/$HOSTNAME"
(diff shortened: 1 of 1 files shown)


Pacman99:
Thanks for the PR! That sounds good to me.

We are actually in the process of completely moving the devos library to https://github.com/divnix/digga. So your change won't have any affect after #276 is merged.

It would be great if you could switch your PR to target digga instead. It should be the same thing except `src/pkgs-lib/shell`.

bb010g:
If #276 will cause `core` to be based off of up-to-date digga, then could this still be merged to `core` here as a backport?

blaggacao:
`develop` is probably going to be merged into `core` in a matter of minutes, so we are on kind of a "freeze" on core at the moment.

Pacman99:
your fix just made it in https://github.com/divnix/devos/commit/634cfcb30162b18f194cb2ca90a54a5c937cba3c

Thank you :)

PR #297, opened 2021-05-24
Added note about cache in start/index.md docs

### LINKED ISSUE #292: Precise in the documentation that one can add the nrdxp cache with sudo if the current user is not trusted
Precise in the documentation that one can add the nrdxp cache with sudo if the current user is not trusted.
This is related to this issue: #287.

Narice:
This resolves #292.
One can use sudo if the user they are using is not trusted by their current system (either their custom NixOS or the NixOS live USB)

### DIFF
--- a/doc/start/index.md
+++ b/doc/start/index.md
@@ -31,6 +31,8 @@ In addition, the [binary cache](../integrations/cachix.md) is added for faster d
 >   files before building the system.
 > - You can choose to simply clone the repo with git if you want to follow
 >   upstream changes.
+> - If the `nix-shell -p cachix --run "cachix use nrdxp"` line doesn't work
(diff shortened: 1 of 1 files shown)


doc/start/index.md
```diff
@@ -31,6 +31,8 @@ In addition, the [binary cache](../integrations/cachix.md) is added for faster d
 >   files before building the system.
 > - You can choose to simply clone the repo with git if you want to follow
 >   upstream changes.
+> - If the `nix-shell -p cachix --run "cachix use nrdxp"` line doesn't work
+>   you can try with sudo: `sudo nix-shell -p cachix --run "cachix use nrdxp"`
```
Pacman99, line 35: This is very explicit, definitely a good thing for documentation on getting started.

Narice: Thanks! and thanks for the merge as well :wink:

PR #300, opened 2021-05-27
flake: update to new digga api

Pacman99:


blaggacao:
Would you agree we should wait the flatten tree / dns implementation and probably also self.inputs instead of inputs?

Pacman99:
> Would you agree we should wait the flatten tree / dns implementation and probably also self.inputs instead of inputs?

Agreed. Don't think this should be merged until digga 0.2.0 is pretty much ready.

flake.nix
```diff
@@ -84,21 +84,31 @@
...
+        importables = rec {
+          profiles = digga.lib.importers.rakeLeaves ./users/profiles;
+          suites = with profiles; rec {
+            base = [ direnv git ];
+          };
         };
       };
 
+      devshell.externalModules = { pkgs, ... }: {
+        packages = [ pkgs.agenix ];
```
blaggacao, line 109: If we want to have the agenix command available in the devshell menu, we'd have to expose it as a `commands = [ { } ];`.

Not sure if that was the intention though, but it might be a good idea, regardless.

blaggacao (commented):
Agenix obersvation ...

flake.nix
Pacman99: Mainly wanted to avoid too much devshell related code in flake.nix. At some point I would like to add a `devshell.toml` in this repo, I just don't have anything to put in there other than agenix.

I would say to add the proper agenix command whenever a devshell.toml is added.

blaggacao: > At some point I would like to add a devshell.toml in this repo
:+1:

Pacman99:
bors try

Pacman99:
bors try

PR #307, opened 2021-06-05
hosts: add DigitalOcean

jwygoda:


Pacman99:
I'm not entirely sure about this change, I don't think we should add host files for such special cases like digitalOcean. I liked the digital ocean image itself, as thats more generic and can work for all hosts.

But adding a host file for it wouldn't really fit with the goals of a "lean" core branch since we'd be adding a host file just for one use case. Most of this stuff a user can figure out with the digital ocean documentation from nixos and devos's deploy-rs docs.

I would be happy to see a documentation addition to `doc/start` maybe `doc/start/digital-ocean.md` that goes over digital ocean setup like how the iso docs go over creating an image and installing it. More documentation on devos special use cases would be great.

blaggacao:
Maybe we can think of ways to integrate with `nixos-generator` project, instead.

PR #325, opened 2021-06-14
feature: Add pkgs fetch manager nvfetcher

### LINKED ISSUE #299: Better package source fetching
Courtesy of research conducted by @GTrunSec

## Would your feature fix an existing issue?

The current way of fetching package sources is by using nix flake builtin capabilities via a sub-flake located in `./pkgs/flake.nix`.

This workflow is seen critically by the community as an overuse of flakes.

https://github.com/berberman/nvfetcher
Implements a flake-native version updater which relies on purpose built tooling for upstream version detection on a variety of source hosting platforms.

It implements a simple to read toml file for specifying those source repositories (instead of sort of "twisting" flake inputs to fit).

It abstracts away the perceived complexity of `./pkgs/flake.nix` entirely.

At some point, flakes might mature into a state where source fetching is natively supported and semantically useful version data (`ref` in addition to `rev`) will be accessible as an input's metadata. However the corresponding suggestion in upstream `nix` (link tbd) was closed by @edolstra for the time being, which indicates that this use case will not enjoy prime support any time soon.

## Additional context

- We have to analyze if the underlying [`nvchecker`](https://github.com/lilydjwg/nvchecker) has superior equal or inferior version fetching qualities than builtin nix.
- To complement the bigger picture within the ecosystem it might also be in order to compare the underlying [`nvchecker`](https://github.com/lilydjwg/nvchecker) sumarily with @ryantm 's `nixpkgs-update`.

GTrunSec:
Regarding motivation in https://github.com/divnix/devos/issues/299

Any question here?

Example: 
-  overlay 
```
inherit (prev.sources.<PackageName>) pname version src;
```

- pkgs 
```
stdenv.mkDerivation rec {
 inherit (sources.<PackageName>) pname version src;
```

Resolves #299


flake.nix
```diff
@@ -46,10 +46,17 @@
...
-            ./pkgs/default.nix
-            pkgs.overlay # for `srcs`
+            #nvfetcher for sources
+            (final: prev:
+              let
+                sources = (import ./sources.nix) { inherit (final) fetchurl fetchgit; };
+              in
+              {
+                inherit sources;
+              })
```
blaggacao, line 56:
```suggestion
            #nvfetcher for sources
            (final: prev: {
              sources = (import ./sources.nix) { inherit (final) fetchurl fetchgit; };
            })
```

flake.nix
```diff
@@ -106,7 +113,14 @@
...
 
       devshell.externalModules = { pkgs, ... }: {
-        packages = [ pkgs.agenix ];
+        packages = [
+          pkgs.agenix
+          pkgs.nvchecker
+          (pkgs.haskellPackages.ghcWithPackages
+            (p: with p;  [
+              nvfetcher
+            ]))
```
blaggacao, line 122: Is this necessary for the Jhon Doe user?

sources.nix
```diff
@@ -0,0 +1 @@
+_build/generated.nix
```
blaggacao, line 1: Can we rather generate `sources.nix` directly? It would be better not to introduce a new top level folder.

nvfetcher.toml
```diff
@@ -0,0 +1,4 @@
+# nvfetcher.toml
```
blaggacao, line 1: Is there a way we can name this file `sources.toml`?

blaggacao:
Super nice! Will continue tomorrow! :rocket:

flake.nix
GTrunSec: sure, if not how to use the command of `nvfetcher build`

nvfetcher.toml
GTrunSec: sure,  we can use nvfetcher build -c `sources.toml`

sources.nix
GTrunSec: This file was generated by nvfetcher.

flake.nix
GTrunSec: Thanks

flake.nix
blaggacao: Oh I had thought the `pkgs.nvchecker` as enough.

sources.nix
blaggacao: Maybe to provide an extension point for manual stuff. Hmm ...

@berberman can you enlighten us?

blaggacao:
We also need to make sure `./_build/.shake*` blobs doen't get committed.

sources.nix
berberman: The feature *extract src* (`extract = [ "file_1", "file_2", ...]`) and *rust support* (`cargo_lock = Cargo.lock`) will extract `file_1`, `file_2`, and `Cargo.lock` to `_build` directory. Extracted files will be referenced by `generated.nix`, and `sources.nix` is a symlink of `_build/generated.nix`. In short, `sources.nix` is no loger the only generated file.

What I have in mind is that we should keep the entire `_build` as the artifact of the build system, but it seems that this is not appropriate in some cases. :thinking:

flake.nix
berberman: I guess we also need `pkgs.nix-prefetch-git` here

flake.nix
```diff
@@ -106,7 +109,14 @@
...
```
berberman, line 118: This is for people who want to use Haskell DSL to define their packages and extend `nvfetcher`, like https://github.com/berberman/flakes/blob/master/Update.hs.

We could append the `buildInputs` of [`ghcWithNvfetcher`](https://github.com/berberman/nvfetcher/blob/b6970c017689ba9495d0c8a64b8d464f0a0aca9f/flake.nix#L32-L38) to this list, so that we don't have to duplicate dependencies of `nvfetcher` again.

sources.nix
blaggacao: Nice feature to start with! :smile:

Can we rather name it `_sources`, then?

flake.nix
blaggacao: I think in this case, we should not place the haskell compiler in the devshell dependencies. Assuming `pkgs.nvfefcher` doesn't have that as a runtime dependency.

GTrunSec: It looks like the `pkgs.nvfecher-bin` enough for us.

sources.nix
GTrunSec: >Can we rather name it _sources, then?

No, we can't do it here. But this should be a PR in `nvfetcher`

nrdxp:
What's the deal with the new issue template? And I'd be really great to have some documentation so I can try this out :laughing:

pkgs/default.nix
```diff
@@ -1 +1,5 @@
-final: prev: { }
+final: prev: {
+  # keep sources this first
+  sources = prev.callPackage (import ./sources.nix) { };
+  # then, call packages with `final.callPackage`
```
blaggacao, line 4: Would this even work? (iirc, it should)

Pacman99: Yes it definitely should

Pacman99: but to use sources in a package, you would probably have to do `final.callPackage`

blaggacao:
Pure cosmetics: https://github.com/berberman/nvfetcher/pull/8

blaggacao:
@berberman Should the shake binary blobs be commited? or is that more like "I'd keep my local track record and that's fine"?

flake.nix
blaggacao: I think [we should be just fine with nvfetcher-bin](https://github.com/berberman/nvfetcher/blob/b6970c017689ba9495d0c8a64b8d464f0a0aca9f/flake.nix#L53)

pkgs/default.nix
blaggacao: I can confirm this works.

overlays/manix.nix
```diff
@@ -0,0 +1,5 @@
+final: prev: {
+  manix = prev.manix.overrideAttrs (o: rec{
+    inherit (prev.sources.manix) pname version src;
```
blaggacao, line 3: using `prev` fails until https://github.com/divnix/digga/commit/cfaad0492cc304d408ab0b4899894623ac363e7a is incorporated. Let's not workaround here. The current code is formally correct.

flake.nix
```diff
@@ -46,10 +46,10 @@
         nixos = {
           imports = [ (digga.lib.importers.overlays ./overlays) ];
           overlays = [
-            ./pkgs/default.nix
-            pkgs.overlay # for `srcs`
             nur.overlay
             agenix.overlay
+            nvfetcher-flake.overlay
+            ./pkgs/default.nix
```
blaggacao, line 52: fixed by https://github.com/divnix/digga/pull/57

&rarr; `./pkgs`

berberman:
> @berberman Should the shake binary blobs be commited? or is that more like "I'd keep my local track record and that's fine"?

It's a persistent state of our build system, which caches some results and helps us produce version changelog. It's fine to keep it locally if we don't need to run `nvfetcher` in CI or other places.

pkgs/sources.nix
```diff
@@ -0,0 +1 @@
+_sources/generated.nix
```
Pacman99, line 1: What is the point of this file? why not just do `prev.callPackage ./_sources/generated.nix`?
The less files added the better.

Pacman99 (approved):
I like this change a lot, thank you for it!

flake.nix
GTrunSec: >We could append the buildInputs of ghcWithNvfetcher to this list, so that we don't have to duplicate dependencies of nvfetcher again.

you are right.

@blaggacao  Sorry,  the nvchecker and nix-prefetch-git are both needed. So `append the buildInputs` would be a solution here.

pkgs/sources.nix
blaggacao: We can't do anything about it it's auto-generated.

Pacman99: `sources.nix` is auto-generated? I thought just _sources/generated.nix was.

blaggacao: Yep

blaggacao:
I'll squash this and open another PR, before w can merge we need to bump digga and incorporate the overlays order fix.

pkgs/sources.nix
Pacman99: thats a little confusing, since its just a symlink. Would be nice to drop it.

blaggacao:
Superseded by https://github.com/divnix/devos/pull/328

nrdxp:
Seems usage is pretty straightforward, but at least a link to the nvfetcher readme would be useful so users are left wondering

Pacman99:
@GTrunSec Just letting you know we updated some functionality and squashed the commits to your branch with a force-push.

GTrunSec:
Ok, Just do it.

blaggacao:
@Pacman99 Can you add the docs commit?

blaggacao (changes requested):
Docs, just to make sure we don't accidentally merge this...

blaggacao (approved):
Thanks! :+1:

Pacman99:
We can send a another commit later to use digga develop once the nixos-generators issues get fixed

pkgs/sources.nix
berberman: Currently the `--no-output` option can disable generating this symlink.

Pacman99 (approved):
@GTrunSec Can you look over the changes and merge if your ok with them(with bors r+)
bors delegate+

GTrunSec:
I got an error output from `nix-shell` with `flake-cmpat` on the latest rev. May you confirm it? Thanks. I guess the problem at here https://github.com/divnix/digga/commit/63538baedbd8bb1fd09687bd1a46eab913a4223b

```
      at /nix/store/p8wl2fqkn99rjmgybgqy4bmjjkw7apy6-source/src/mkFlake/default.nix:85:23:

           84|
           85|     devshellModules = lib.exporters.modulesFromList {
             |                       ^
```

blaggacao:
Oh, thnaks! [`develop` has the fixup](https://github.com/divnix/digga/commit/88f0431746859f09e4b4db4fafb4e9b57ed7e23c)

GTrunSec:
Now, We can commit some change to develop digga for flake check passing.
bors delegate+

flake.nix
```diff
@@ -48,7 +48,9 @@
           overlays = [
             nur.overlay
             agenix.overlay
-            nvfetcher-flake.overlay
+            (final: prev: {
+              nvfetcher-bin = nvfetcher-flake.defaultPackage.${final.system};
```
GTrunSec, line 52: I got an issue with `nvfecther-flake.overlay`

```
 attribute 'nvfetcher' missing

       at /nix/store/hf8zm40wwi03wwmjd88djhzhkw49qqbn-source/flake.nix:60:47:

           59|           nvfetcher-bin = with final;
           60|             haskell.lib.justStaticExecutables haskellPackages.nvfetcher;
             |                                               ^
```

flake.nix
```diff
@@ -69,7 +71,7 @@
         hostDefaults = {
           system = "x86_64-linux";
           channelName = "nixos";
-          modules = ./modules/module-list.nix;
+          imports = [ (digga.lib.importers.modules ./modules) ];
```
GTrunSec, line 74: Broken target:
```
attempt to call something which is not a function but a list                                                      
                                                                                                                         
       at /nix/store/f2wp2m962hvb620vmz5yc6fi794ijqz6-source/lib/modulesFromList.nix:39:17:                              
                                                                                                                         
           38|       hasFileAttr = o: builtins.hasAttr "_file" o;                                                        
           39|       peek = f: f (builtins.functionArgs f);                                                              
             |                 ^                                                                                         
           40|                                       
```

blaggacao: Thanks! I'll klean my mess up tomorrow. 😀

pkgs/default.nix
```diff
@@ -1,5 +1,6 @@
-final: prev: {
+inputs: final: prev: {
+  nvfetcher-bin = inputs.nvfetcher-flake.defaultPackage.${final.system};
```
Pacman99, line 2: Lets not pass `inputs` to files, thats what `overlays` is for, can we instead just add an overlay that adds, nvfetcher's defaultPackage

flake.nix
```diff
@@ -21,20 +21,20 @@
       agenix.inputs.nixpkgs.follows = "latest";
       nixos-hardware.url = "github:nixos/nixos-hardware";
 
-      pkgs.url = "path:./pkgs";
-      pkgs.inputs.nixpkgs.follows = "nixos";
+      nvfetcher-flake.url = "github:berberman/nvfetcher";
```
Pacman99, line 24: Also can this just be `nvfetcher`, we haven't added `-flake` to any other input

pkgs/default.nix
Pacman99:
```nix
overlays = [
  (final: prev: { nvfetcher = nvfetcher.defaultPackage.x86_64-llinux; })
];
```
in flake.nix

flake.nix
Pacman99: I noticed this error, it seems like it might be a problem in upstream

blaggacao:
I guess we should bump digga independently of this PR.

pkgs/default.nix
GTrunSec: thanks.

GTrunSec:
>I guess we should bump digga independently of this PR.

agree with that.

flake.nix
```diff
@@ -111,7 +111,8 @@
           {
             name = pkgs.nvfetcher-bin.pname;
             help = pkgs.nvfetcher-bin.meta.description;
-            command = "cd $DEVSHELL_ROOT/pkgs; ${pkgs.nvfetcher-bin}/bin/nvfetcher -c ./sources.toml $@"; }
+            command = "cd $DEVSHELL_ROOT/pkgs; ${pkgs.nvfetcher-bin}/bin/nvfetcher -c ./sources.toml --no-output $@; nixpkgs-fmt _sources/";
```
GTrunSec, line 114: nixpkgs-fmt _sources/  for cleanup the trailing whitespace

Pacman99:
bors try

blaggacao (changes requested):
Unfortunately upstream is broken, so I'm blocking this until I could resolve it.

nrdxp:
upstream nixpkgs, or digga?

blaggacao:
> digga?

https://github.com/divnix/digga/pull/60

I've fixed `fup` in the mean time, where I had implemented `modulesFromList` working with `_file` attrs to short-sightedly. This change then surfaced https://github.com/divnix/digga/pull/60

nrdxp:
bors try

nrdxp:
Really would like to get this sucker merged. After both of your above mentioned PR's have been merged above @blaggacao, what's left blocking this? Do we just need to bump digga?

blaggacao:
@GTrunSec can you preemtively rebase on https://github.com/divnix/devos/pull/329 ? or maybe we can merge it before you read this, whatever comes first...

GTrunSec:
Yes, we can merge it. Can you do that?

GTrunSec:
bors try

nrdxp (approved):
bors r+

blaggacao (approved):

blaggacao:
bors r+

(everything else will be fixup &mdash; if so)

nrdxp:
:crossed_fingers:

PR #326, opened 2021-06-14
ref: pass our lib as lib.our to the module system

blaggacao:
At least we don't find ourselves with strange module signatures
and use something that looks like a dedicated namespace within
he module system: `config.lib`

---

```shell
➜  devos git:(da/pass-lib-as-config-lib) rg ourLib
➜  devos git:(da/pass-lib-as-config-lib) # nothing found
```


Pacman99 (approved):
yup this looks much cleaner
bors r+

PR #334, opened 2021-06-30
ref: move devshell module here to `devos` again

blaggacao:
This is tested via https://github.com/divnix/devos/pull/334

If that is green, we can merge there update the ref to `digga` here and run the CI here again.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,11 @@
       digga = {
         url = "github:divnix/digga/develop";
         inputs.nipxkgs.follows = "latest";
+        inputs.deploy.follows = "deploy";
       };
+      bud.url = "github:divnix/bud"; # no need to follow nixpkgs: it never materialises
+      deploy.url = "github:serokell/deploy-rs";
+      deploy.inputs.nixpkgs.follows = "nixos";
 
       ci-agent = {
         url = "github:hercules-ci/hercules-ci-agent";
@@ -31,6 +35,7 @@
   outputs =
     { self
     , digga
+    , bud
     , nixos
     , ci-agent
     , home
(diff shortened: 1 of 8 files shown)


shell/flake.nix
```diff
@@ -0,0 +1,63 @@
...
+
+  in
+  {
+
+    overlays = {
+      enable-deploy = deploy.overlay;
+      patched-nix = import ../patchedNix;
+    };
+
+    devShell = ufrContract supportedSystems ./. flkShellInputs;
```
Pacman99, line 48: It doesn't make sense to me to add the ufr polyfills here since digga is the one that does all the exporting. I think it would be confusing and unnecessary to  add it here.

blaggacao: `flake.nix` & `flake.lock` are now removed: it was a sort of leftover from the refactoring journey, where at some stage, before outfactoring `flk`, the `flkShell` had the character of a flake of its own right and domain.

Pacman99:
Can we move the `bud` folder under `shell` to spare the extra top-level folder. Lets to keep a clean root folder.

blaggacao:
> Can we move the bud folder under shell to spare the extra top-level folder. Lets to keep a clean root folder.

I hear the pain, but unfortunately that would be a domain violation: `bud` _can_ be used for shell, but actually is more like a special `pkgs`. I'll evaluate if we can move it there, instead...

flake.nix
```diff
@@ -31,15 +38,20 @@
...
     , home
     , nixos-hardware
     , nur
     , agenix
     , nvfetcher
+    , deploy
     , ...
     } @ inputs:
+    let
+      bud' = bud self; # rebind to access self.budModules
```
Pacman99, line 53: I really don't like this line, I hope we can remove it soon

Pacman99 (approved):

flake.nix
blaggacao: Ok, going to do a refactoring ASAP taking this into account. Thanks!

blaggacao:
bors try

PR #340, opened 2021-07-14
imp: add portable user definition

blaggacao:
According to https://github.com/nix-community/home-manager/issues/2161
there might be a better formal distinction between host-spaced and
portable arch-spaced hm users in the future.


blaggacao:
bors try

blaggacao:
bors try

PR #348, opened 2021-07-25
Automatically build sources

Th3Whit3Wolf:
This pull request does a couple things.

1. Any package inside `./pkgs` (with the exception of `./pkgs/bud`, `./pkgs/_sources` and `./pkgs/default.nix` will be called (as in `callPackage`) and can use `inherit (source) pname version src` so long as the name of parent folder for the `default.nix` or the name of the nix file stripped of .nix is the same name used in `sources.toml`
2. Start work on automatically building vimPlugins, vscode extension, and other development tools from prefixes in `sources.toml` Currently only vimPlugins are implemented.

I recognize that `flake.nix` is not a great place for an overlay. I was unable to find a way to make an overlay that lets you import inputs, which means defining rakeTree and flattenTree before writing what's currently in the flake. It would also have had to be used as `(import ./path/filename.nix).overlay`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,6 +63,36 @@
     , deploy
     , ...
     } @ inputs:
+
+    let
+      rakePkgs = dir:
+        let
+          sieve = name: val:
+            (name != "default" && name != "bud" && name != "_sources");
+
+          flattenFiltered = digga.lib.flattenTree
+            (nixos.lib.filterAttrs sieve (digga.lib.rakeLeaves dir));
+          getBasename = name: nixos.lib.last (nixos.lib.splitString "." name);
+        in
+        nixos.lib.mapAttrs' (n: v: nixos.lib.nameValuePair (getBasename n) v)
+          flattenFiltered;
+
+      localPackages = final: prev:
+        builtins.mapAttrs
+          (name: value:
+            let
+              sources = (import ./pkgs/_sources/generated.nix) {
+                inherit (prev) fetchurl fetchgit;
+              };
+              package = import (value);
+              args = builtins.intersectAttrs (builtins.functionArgs package) {
+                source = sources.${name};
+              };
+            in
+            final.callPackage package args)
+          (rakePkgs (./pkgs));
+
+    in
     digga.lib.mkFlake
       {
         inherit self inputs;
@@ -78,6 +108,7 @@
               agenix.overlay
               nvfetcher.overlay
               deploy.overlay
+              localPackages
               ./pkgs/default.nix
             ];
           };
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -1,5 +1,29 @@
-final: prev: {
-  # keep sources this first
-  sources = prev.callPackage (import ./_sources/generated.nix) { };
-  # then, call packages with `final.callPackage`
+self: super:
+let
+  sources = (import ./_sources/generated.nix) { inherit (self) fetchurl fetchgit; };
+
+  mkVimPlugin = plugin:
+    self.vimUtils.buildVimPluginFrom2Nix {
+      inherit (plugin) pname version src;
+    };
+
+  newPkgsSet = pkgSet:
+    let
+      prefix = "${pkgSet}-";
+
+      pkgSetBuilder = {
+        "vimPlugins" = mkVimPlugin;
+      }.${pkgSet};
+
+
+      pkgsInSources = self.lib.mapAttrs' (name: value: self.lib.nameValuePair (self.lib.removePrefix prefix name) (value)) (self.lib.filterAttrs (n: v: self.lib.hasPrefix prefix n) sources);
+    in
+    self.lib.mapAttrs (n: v: pkgSetBuilder v) pkgsInSources;
+
+in
+{
+  inherit sources;
+
+  vimPlugins = super.vimPlugins // (newPkgsSet "vimPlugins");
+
 }


flake.nix
```diff
@@ -63,6 +63,36 @@
...
+        let
+          sieve = name: val:
+            (name != "default" && name != "bud" && name != "_sources");
+
+          flattenFiltered = digga.lib.flattenTree
+            (nixos.lib.filterAttrs sieve (digga.lib.rakeLeaves dir));
+          getBasename = name: nixos.lib.last (nixos.lib.splitString "." name);
+        in
+        nixos.lib.mapAttrs' (n: v: nixos.lib.nameValuePair (getBasename n) v)
+          flattenFiltered;
```
blaggacao, line 78: So you do think an `importPackages` would be suitable addition to `digga`? Could be, but we'd have to make and discuss the use case [over there](https://github.com/divnix/digga).

pkgs/default.nix
```diff
@@ -1,5 +1,29 @@
-final: prev: {
-  # keep sources this first
-  sources = prev.callPackage (import ./_sources/generated.nix) { };
-  # then, call packages with `final.callPackage`
+self: super:
```
blaggacao, line 1: In flake world those are by convention: `final: prev:`.

That this is not enforced by the flake type checker here, is more a coincidence than a reason to deviate, i.m.o.

flake.nix
Th3Whit3Wolf: Yes I do, I use the rakePkgs function exclusively to build the localPackages overlay. Which builds all packages in the `./pkgs` directory so I never have to touch `./pkgs/default.nix` when I add a new package. I do not know how to move it into digga and function properly.

blaggacao: we could make `pkgs` complely `default.nix`-less and add add `nvfetcher` stuff separately.

We could also consider to teach `rakeLeaves` to not consider basenames starting with an underscore.

blaggacao: Maybe you can have a quick look at the digga impl. With a little help, it shouldn't be too hard.
https://github.com/divnix/digga/blob/main/src/importers.nix

Th3Whit3Wolf: I tried actually, rakePkgs is easy but I could not figure out how to make the localPackages overlay fit in there.

blaggacao: I'd model it after https://github.com/divnix/digga/blob/fb3497d6d1ee28abc61170b52369ffca423681a5/src/importers.nix#L120-L127

returning a module that can be `imports`ed:
```nix
{
  overlays = {};
}
```

blaggacao: I think you can use `rakeLeaves` as is. the only (orthogonal) addition could be to skip on `_` prefixes.

blaggacao: In theory, you could also reuse `flattenTree` and derive the appropriate package name from that string, and finally `callPackage` on its values.

blaggacao:
now, I got a clearer picture what this is all about :smile: 

@danielphan2003 Can `vs-ext` generalize well enough over `<package set>` so that we could actually inaugurate a pkgsSet lib? Or merge anything +- viable back into `digga`?

danielphan2003:
Yes, it can. Head over to [`vs-ext`](https://github.com/divnix/vs-ext) for usage, assuming this PR is merged.

blaggacao:
I kind of see a `divnix/vim-ext` repo not too far on the horizon. @Th3Whit3Wolf what do you think?

It would make more sense to be able to iterate independently of the `devos` release cycle. :rocket: 

Relatively soon (`v1.0`) `devos` will become very boring and slow motion.

Th3Whit3Wolf:
> I kind of see a divnix/vim-ext repo not too far on the horizon. @Th3Whit3Wolf what do you think?

We could do that, would it be better to have a repo for each `<package set>` or one repo that has all of them?

blaggacao:
> would it be better to have a repo for each <package set> or one repo that has all of them?

That depends probably on two things:

How much design overlap there is and how much does it hurt if release cycles have to be coupled.

Do you have any idea about what we could expect?

Th3Whit3Wolf:
I think multiples repos are best for now.

I think there could also be a future where one larger repo or maybe even digga creates the overlay itself as to not clutter the users' flake inputs.

blaggacao:
https://github.com/divnix/vim-ext the project is yours :rocket:

Th3Whit3Wolf:
@blaggacao sorry it's taken me so long to respond, I've been quite busy lately. 

I've been looking at [vs-ext](https://github.com/divnix/vs-ext) and I'm not sure it's best to make vim-ext right now. Unless I'm mistaken the goal of vs-ext updates is to take use bud to take a vscode extension unique id, query openvsx, and then update that extensions metadata. 

This **_could_** be done for vim plugins however it would be a lot more work. Vim plugins aren't collectively stored in one location. You can find them on github,gitlab,(any code hosting service), [vimawesome](https://vimawesome.com/), and [nvim-awesome](https://nvim-awesome.vercel.app/) So there would be need to create a method to query each one. If vim-ext is being described as 

> A kick ass library to dominate your Vim Extensions (with DevOS) 

I would expect it to build vim plugins that have build requirements (could be node, rust, clojure, or really anything).

Th3Whit3Wolf:
Follow up question, how much reliant is devos on nvfetcher?  

I've started playing around with something similliar to nvfetcher with a few key differences 

1. you don't have to declare fetch (just src)
2. My intentions are to generate overlays, with metadata, using platform apis (github, gitlab, openvsx, vscode).

Pacman99:
> Follow up question, how much reliant is devos on nvfetcher?

We made a general decision to use nvfetcher over flakes for updating srcs because it was a cleaner solution that was built for the use case. But if another solution is better, I'm sure minds can be changed :). We would have to run through a similar discussion/debate again.

blaggacao:
@Th3Whit3Wolf this time sorry for me chiming in late for the same reason as yours.

Well, one day, if subflakes get really advanced, we might even want to switch something flake compatible, since there is a benefit in creating a SBOM easily from the `flake.lock`.

`vim-ext` can be also really small. The point in having it in a repo also is a bit _not_ having it in `devos` or `digga` (for now). The reason is that all those editors might a) either have their own dynamic and benefit from a separate release schedule for the users (`nix flake lock --update-input vim-ext`). Another reason is that beyond some initial preferences of Tim, we would probably try to avoid being opinionated with anything that doesn't directly relate to flakes/nix. So by making it a separate repo, we essentially avoid "pissing of" those who don't use X or Y. A third reason is maintenance and support. If the project grows, then it will be helpful to forward `vim-ext` issues to the repo in order to create "tighter" feedback cycles between authors and users and convert users more easily into co-authors / co-maintainers once they finde a tiny and easy to understand codebase. `vs-ext` / `vim-ext` can well be the 'entry drug' heheheh. :smile:

PR #352, opened 2021-08-01
Chore

GTrunSec:
We don't need the nixpkgs-fmt for nvfetcher anymore, and it has been fixed upstream.

### DIFF
--- a/bud/default.nix
+++ b/bud/default.nix
@@ -6,5 +6,11 @@
       help = "Copy the desired template to DEST";
       script = ./get.bash;
     };
+    nvfetcher-github = {
+      writer = budUtils.writeBashWithPaths [ nvfetcher-bin coreutils git ];
+      synopsis = "nvfetcher-github";
+      help = "Auto update with nvfetcher on github action";
+      script = ./nvfetcher-github.bash;
+    };
   };
 }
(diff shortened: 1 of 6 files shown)


blaggacao:
bors try

GTrunSec:
What do you think about the script of auto-update with nvfetcher on github action
Example: https://github.com/GTrunSec/vast2nix/commit/f17dcce62bee61e5ce2a304c5ff1d033ba387c31

blaggacao:
Nice! [`8f1a70f` (#352)](https://github.com/divnix/devos/pull/352/commits/8f1a70f641abd596705b32627c1a309a2c6d6cdd)

Is the [`sources.toml` literal](https://github.com/divnix/devos/pull/352/commits/8f1a70f641abd596705b32627c1a309a2c6d6cdd#diff-3c5e4fb56e10492f855fb66b7416ba525ce912d0ecdd4dad3c9da4877e3be260R2) the only reason why this is not generic enough to go into `bud`'s `stdProfile.nix`? If we can make it generic, it might make even more sense there?

It also looks like it could be made generic in a sense that it can be used either as a github action or from the command line. w.t.y.t. ?

blaggacao:
chores cherry picked in `main`. Let's move the `bud` tooling to `bud`.

Issue #369, opened 2021-09-18
Primary branch name in CI configurations points to `main` but `main` was renamed to `master`


montchr:
## Expected Behavior

CI workflow configurations and documentation should reference the current primary branch. At this point in time, that is `master`, but ideally `master` would be removed in favor of a  `main` branch as the primary branch on GitHub.

## Current Behavior

There are still [a few references](https://github.com/divnix/devos/search?q=main) throughout the project to `main`, which appears to have been renamed to `master` (see #368).

The primary result of this is that CI workflows are not running upon pushes to `master` as they had been for `main`.

## Possible Solution

Rename `master` to `main` if currently feasible.

If not currently feasible, replace references to `main` with `master` to reflect the repo's current state until `master` can be renamed to `main` in line with conventions and the previous state of the repo.

## Steps to Reproduce

Most notably `main` is left over in some CI configurations:

https://github.com/divnix/devos/blob/a3b2bcf3d57be28c27119a44445cfdde38349cb9/.github/workflows/mdbook_docs.yml#L1-L6

https://github.com/divnix/devos/blob/a3b2bcf3d57be28c27119a44445cfdde38349cb9/.github/workflows/check.yml#L1-L7

`main` is also referred to as the preferred base branch for pull requests in https://github.com/divnix/devos/blob/master/doc/CONTRIBUTING.md 

## Context

I'm guessing this explains not only why the live docs site is out of sync with `master` but also why checks and Cachix update workflows haven't been running.

It might also be the cause of #363 since 079adc4474231d5582fee5574bc5bcc6f133e5ac was the last commit to cause that workflow run.

blaggacao:
Great Analysis! Thx!

montchr:
Sounds good, thank you. I think #371 sufficiently addresses the issue in that case. Actions are running again! I'll close this out.

PR #371, opened 2021-09-18
Trigger CI workflows on pushes to both `main` and `master` + Update PR target branch guideline

montchr:
Addresses the broken CI workflows mentioned in #369 by allowing workflow triggers on pushes to the `master` branch since `main` doesn't exist. This does not necessarily mean #369 can be closed since ideally `master` will be replaced with `main` entirely.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - master
       - trying
       - staging
 jobs:
--- a/.github/workflows/mdbook_docs.yml
+++ b/.github/workflows/mdbook_docs.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - master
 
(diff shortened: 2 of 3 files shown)


blaggacao:
Thank you a ton! I'll promise to merge any such cleanup PR coming in within the next week within 24 hours.

This is really long overdue and you are doing new folks a great service! (To not to get confused)

montchr:
Thanks!

I'm gradually working my way through the process of getting up and running with the DevOS template, molding my own configuration into a shape more similar to that of the template. I hope to provide some more feedback/updates as I progress if I run into other points of confusion that seem like they might be useful to others (rather than the confusion resulting from my own inexperience with NixOS) 😁

I appreciate all the effort all of you have spent on this project 🙂 – it's refreshing to have such a reference point for what can be done with flakes and reusable Nix configurations, even as the template and the API are in flux.

Issue #381, opened 2021-10-01
Renaming default user


blm768:
I'm playing around with a basic DevOS configuration, and I'd like to rename the default user, but something seems to be hard-coded to refer to the `nixos` user. Here's what I've attempted so far:

1. Start from commit `287cb82d`.
2. Run `bud vm NixOS`. The VM builds.
3. Apply the following patch:

   ```patch
   diff --git a/flake.nix b/flake.nix
   index d87fdce..dfb458c 100644
   --- a/flake.nix
   +++ b/flake.nix
   @@ -123,7 +123,7 @@
                  users = digga.lib.rakeLeaves ./users;
                };
                suites = with profiles; rec {
   -              base = [ core users.nixos users.root ];
   +              base = [ core users.test users.root ];
                };
              };
            };
   diff --git a/hosts/bootstrap.nix b/hosts/bootstrap.nix
   index d491855..04179ec 100644
   --- a/hosts/bootstrap.nix
   +++ b/hosts/bootstrap.nix
   @@ -8,7 +8,7 @@
        # profiles.networking
        profiles.core
        profiles.users.root # make sure to configure ssh keys
   -    profiles.users.nixos
   +    profiles.users.test
      ];
    
      boot.loader.systemd-boot.enable = true;
   diff --git a/users/nixos/default.nix b/users/test/default.nix
   similarity index 56%
   rename from users/nixos/default.nix
   rename to users/test/default.nix
   index 5c037fe..daed888 100644
   --- a/users/nixos/default.nix
   +++ b/users/test/default.nix
   @@ -1,10 +1,10 @@
    { hmUsers, ... }:
    {
   -  home-manager.users = { inherit (hmUsers) nixos; };
   +  home-manager.users = { inherit (hmUsers) test; };
    
   -  users.users.nixos = {
   +  users.users.test = {
        uid = 1000;
   -    password = "nixos";
   +    password = "test";
        description = "default";
        isNormalUser = true;
        extraGroups = [ "wheel" ];
   ```
4. Now `bud vm NixOS` fails with the following:
   ```
   error: attribute 'test' missing

       at /nix/store/37v2bwx9ij1gnm226v7k3ydjm4knz0j2-source/users/test/default.nix:3:25:

            2| {
            3|   home-manager.users = { inherit (hmUsers) test; };
             |                         ^
            4|
   (use '--show-trace' to show detailed location information)
   ```

Reverting the change to the `home-manager.users` line produces a different error:
```
error:
[...]

PR #385, opened 2021-10-04
Fix outdated option name

blm768:
While muddling through #381, I realized the docs had a reference to an option that had been renamed.

I'm not sure yet whether any changes should be made to the example in the previous section, as the default seems to be to pull stuff in from `hmUsers` rather than directly specifying options like the example does, but I'm not sure what we'd consider best practice there.

### DIFF
--- a/doc/concepts/users.md
+++ b/doc/concepts/users.md
@@ -37,9 +37,10 @@ User profiles can be collected in a similar fashion as system ones into a `suite
 argument that gets passed to your home-manager users.
 
 ### Example
+`flake.nix`
 ```nix
 {
-  home-manager.users.nixos = { suites, ... }: {
+  home.users.nixos = { suites, ... }: {
     imports = suites.base;
   };
 }


Pacman99:
Wow I'm pretty sure that option has always been named `home`, I can't believe I missed that.
Thanks for catching this!

Pacman99:
Ohh I see I think we need be more clear as to what we're referring to, but the option is actually `home-manager`. That code is using the [home-manager nixos module](https://github.com/nix-community/home-manager/blob/master/nixos/default.nix), not the digga api.

blm768:
> That code is using the home-manager nixos module, not the digga api.

Right. I think the example should work as given, but it doesn't match the current state of the template, which presumably expects the profiles to be configured in `flake.nix`:

```nix
{
  home-manager.users = { inherit (hmUsers) nixos; };
}
```

blm768:
> Ohh I see I think we need be more clear as to what we're referring to, but the option is actually `home-manager`. That code is using the [home-manager nixos module](https://github.com/nix-community/home-manager/blob/master/nixos/default.nix), not the digga api.

Were you referring to the option I changed or the option in the example above it? It's been a while now, but I think I remember that the one I changed was using `digga`'s API.

doc/concepts/users.md
```diff
@@ -39,7 +39,7 @@ argument that gets passed to your home-manager users.
 ### Example
```
Pacman99, line 39:
```suggestion
### Example
`flake.nix`:
```

Pacman99:
Right yeah you were correct originally, I'm sorry about my confusion.
I just added a comment with a slight change to specify what file the docs are talking about. With that, I'm happy to merge this.

PR #395, opened 2021-10-24
Remove git crypt, update docs

b12f:
`git crypt` is rather hard to use, and overlaps with `agenix` in a lot of circumstances. Using both of them at the same time can result in double-encrypted files, causing issues that are hard to debug. This PR removes git-crypt from the installation and documentation, favoring agenix instead.


nrdxp:
I think it's time we go ahead and pull the trigger on this.

PR #407, opened 2021-12-29
profiles/core: add comments for each option

### LINKED ISSUE #295: Core profile optional (doc)
Let's document via well contextualized code comments how the different parts of `core` are optional.

The only thing that has a drawback would probably be openssh disabled because no host cert is autocreated.

/cc @Radvendii

Pacman99:
closes #295

### DIFF
--- a/profiles/core/default.nix
+++ b/profiles/core/default.nix
@@ -2,12 +2,21 @@
 let inherit (lib) fileContents;
 in
 {
+  # Sets nrdxp.cachix.org binary cache which just speeds up some builds
   imports = [ ../cachix ];
 
+  # For rage encryption, all hosts need a ssh key pair
+  services.openssh = {
+    enable = true;
+    openFirewall = lib.mkDefault false;
+  };
+
+  # This is just a representation of the nix default
   nix.systemFeatures = [ "nixos-test" "benchmark" "big-parallel" "kvm" ];
 
   environment = {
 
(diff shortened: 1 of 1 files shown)


profiles/core/default.nix
```diff
@@ -2,12 +2,21 @@
 let inherit (lib) fileContents;
 in
 {
+  # Sets nrdxp.cachix.org binary cache which just speeds up some builds
   imports = [ ../cachix ];
 
+  # For rage encryption, all hosts need a ssh key pair
```
Pacman99, line 8: Moving this to the top, since it is somewhat required for agenix, so I want to emphasize the importance

profiles/core/default.nix
```diff
@@ -113,18 +124,18 @@ in
...
     autoOptimiseStore = true;
-
     gc.automatic = true;
-
     optimise.automatic = true;
 
+    # Prevents impurities in builds
     useSandbox = true;
 
-    allowedUsers = [ "@wheel" ];
```
Pacman99, line 124: Removing this, allowedUsers breaks systemd services that want to run nix, which I think a couple do. In terms of security, i believe its ok to let any user run nix since they can't really affect system state.

blaggacao (approved):

Pacman99:
Going to manually merge, CI is not currently working and this is just a docs change

PR #414, opened 2022-01-09
Update secrets path

blm768:
The path for decrypted secrets has changed upstream.


Pacman99:
Ohh right thanks for updating that!

PR #420, opened 2022-01-29
Switch back to pulling devos as a template

Pacman99:
With divnix/digga#121, the infinite recursion issue when devos is not a git repository should be fixed. Pulling devos as a template is the correct method of getting devos, so lets switch back to doing that.


Pacman99:
bors try

Pacman99:
combining with #421

PR #422, opened 2022-01-30
Update inputs and switch back to pulling devos as a template

Pacman99:


Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

blaggacao (approved):

blaggacao:
bors +

Pacman99:
I think when it tries to pull some hm file from the cache its failing. The "unexpected end of file" indicates a cache issue according to a nix issue. Which is why it works the first time when that files gets built, but not the second time when it gets pulled from the cache.

Pacman99:
Going to merge since the build succeeded the first time and nothing has changed since then. We should try to figure out whats going on with the cache configuration and the CI

PR #427, opened 2022-02-18
flake: switch back to serokell deploy input

Pacman99:
There is a bug in the input-output-hk fork

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -102,7 +102,6 @@
     },
     "deploy": {
       "inputs": {
-        "fenix": "fenix",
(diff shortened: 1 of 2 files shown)


nrdxp (approved):
IIRC, the fork doesn't even function properly as a standalone cli tool anymore, since it was integrated into the bitte binary. Perhaps I am misunderstanding though, but lgtm so far.

Pacman99:
It gives an error about being unable to create a socket for the hostname

blaggacao:
Ah, this is cause it requires a port specified, yes. This is known, indeed.

Issue #430, opened 2022-02-23
Merge into digga


Pacman99:
Maintaining two separate repositories for such connected projects is becoming increasingly difficult. Anytime something changes in digga, an accompanying pull request has to be made in devos to fix the changes. We're maintaining two "separate" docs and projects when they really are the same thing.

I think this repo should just be a folder in the digga examples folder instead of being its own repo. The majority of issues in this repo are really about digga. Most of the "devos" docs are just about applications of the digga api.

Merging this repo into digga wouldn't change a whole lot. We can export the devos template with the `templates` output in digga, so all the getting started docs will still apply, but just with the digga api instead. `nix flake new -t github:divnix/digga`.

In terms of discoverability, we can move all the devos docs to be part of digga's docs and link the getting started in the digga readme.

cc @nrdxp @blaggacao

blaggacao:
Perfect!

In order to.preserve github's ⭐, we might need to do a bit of a rochade:

- rename digga to something else

- merge digga into devos the way you propose

- rename devos to digga


---

This is really a simple, though geniouy idea on how to solve the maintenance issue.

Pacman99:
Great idea! That way we can also preserve issues on devos too

Pacman99:
Going to pin this issue, so people know what just happened

Pacman99:
For contributors, its going to be awkward since this repository's forks are still devos. I think you can just delete your digga fork and rename your devos fork to digga then merge in main.

But if you have important branches in your digga fork, make sure to copy those over. Based off current pull requests in digga, this should mostly be a problem for @blaggacao @montchr and me.

nrdxp:
for the record, I was against the split in the beginning :trollface:

PR #431, opened 2022-02-23
Make this repository digga

Pacman99:
closes: #430

This is just some git hacking to preserve both project's commit histories.

Commands that I ran:
```sh
git remote add digga git@github.com:divnix/digga
git checkout -b digga-main digga/main
git checkout -b merge-digga upstream/main
mkdir ../devos
mv * .editorconfig .envrc .gitignore ../devos/
mkdir examples
mv ../devos examples/
git add . && git commit -m "tree: move everything into examples/devos"
git checkout digga-main
git rebase merge-digga # there were 3 merge conflicts with src/pkgs-lib/shell, not sure why but that file has been deleted so I just delt with them
git checkout merge-digga
git merge digga-main --ff-only
git push upstream merge-digga
```
There are a lot of changes, so it'll be easier to review what exactly I did with git to get it this way.

The only actual changes I personally made:
 - removed .gitmodules and downstream submodule
 - move devos into examples/devos
 - removed bud template from devos and added `devos` template to digga

This doesn't cover devos docs, I think we will need a separate effort to merge devos docs into digga docs. Probably as a digga usage section or something. For now devos docs are just sitting in `examples/devos`.

This PR should theoretically have the same amount of commits being added as the number of commits in digga itself. I think it ended up being around 50 less because bors commits aren't included for some reason.


Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
bors try

Pacman99:
> I can try tomorrow to do a git filter-repo merge of two unrelated histories whereby the devos history lands entirely in examples/devos, albeit all commits will be rewritten with the new subdir prefix.
> 
> Then, I can replay the manual changes that you did on top.

I don't think that's a good idea because it would rewrite this repos git history which is never good for public branches.  It would make sense if were merging devos into digga, but not when were doing it this way.

blaggacao:
Ah! I see. i think I got lost first pass replaying your steps in my head. I then came ubder the impression commit history was lost, even if you explicitly stated otherwise. :shrug:

blaggacao:
bors r+

In this case rather wake up with a unified Digga/DevOS.

Thank you!

PR #435, opened 2022-02-26
Merge devos docs and README into top-level

Pacman99:
I think some more work needs to be done to get the docs to flow better, but for now this puts everything in one place so we have a starting point to improve everything.

The new readme is a mix of what I thought were the best parts of both readmes.


blaggacao (approved):
Good strating point, indeed. Thank you! :rocket:

PR #436, opened 2022-02-26
Remove customBuilds, nixos-generate has --flake

Pacman99:
nixos-generators now has --flake support, so theres no reason for us to export all the builds here. Lets just offload this responsibility to nixos-generators.

Somewhat blocked by nix-community/nixos-generators#138, so that we don't need to make our hosts overridable.

### DIFF
--- a/examples/devos/shell/devos.nix
+++ b/examples/devos/shell/devos.nix
@@ -45,10 +45,11 @@ in
     (docs mdbook)
     (devos inputs.deploy.packages.${pkgs.system}.deploy-rs)
   ]
-
   ++ lib.optional
     (system != "i686-linux")
     (devos cachix)
-
+  ++ lib.optional
+    (system != "aarch64-darwin")
+    (devos inputs.nixos-generators.defaultPackage.${pkgs.system})
   ;
 }
--- a/examples/devos/flake.lock
+++ b/examples/devos/flake.lock
@@ -369,6 +369,21 @@
         "type": "github"
       }
     },
+    "nixlib_2": {
+      "locked": {
+        "lastModified": 1636849918,
+        "narHash": "sha256-nzUK6dPcTmNVrgTAC1EOybSMsrcx+QrVPyqRdyKLkjA=",
+        "owner": "nix-community",
+        "repo": "nixpkgs.lib",
+        "rev": "28a5b0557f14124608db68d3ee1f77e9329e9dd5",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "nixpkgs.lib",
+        "type": "github"
+      }
+    },
     "nixos": {
       "locked": {
         "lastModified": 1643463207,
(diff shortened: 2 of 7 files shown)


montchr:
Nice. I had to work around `customBuilds` in the `darwin-support` branch since it's not usable on darwin hosts. I was going to propose via that branch that we make the module optional or loaded only for Linux hosts, but offloading it to `nixos-generators` is even better.

montchr (approved):
🏗️

danielphan2003:
I used to use `bud` to generate ISOs for my configs, but now that https://github.com/nix-community/nixos-generators/pull/138 just closed, how do I use nixos-generate with this?

I'm using this PR and the latest commit of nixos-generators. I used `nixos-generate -f iso --system x86_64-linux --flake .\#rog-bootstrap` but it throws `attribute 'extendModules' missing'`. Does this have anything to do with the old `bud` behavior (see https://github.com/divnix/bud/blob/a789d710851441ba7e7cd59be378623b1fe05688/scripts/hosts-build.bash)?

Pacman99:
The `extendModules` error is likely from your host being evaluated with an older version of the module system, so you would have to update to 21.11 or after. If you are already updated, then I'm not really sure whats wrong.
This is the the commit that added `extendModules ` https://github.com/NixOS/nixpkgs/pull/143207, so can you check to see if you have that commit. You can also investigate in nix repl, to see if `extendModules` is in your config.

I was able to build an iso with my host, and it seems like it includes my entire hosts closure in the iso. I think the `install-iso` format feasibly replaces our `bootstrapIso` with everything but the networking part. If that was to be upstreamed to nixos-generators, then we could remove `bootstrapIso` and `isoConfig`.

danielphan2003:
I can only find `Flake.outputs.nixosConfigurations.<host>._module.args.extendModules`, but not top-level extendModules. I forked https://github.com/danielphan2003/nixos-generators/blob/module-args-extendModules/nixos-generate.nix to use `_module.args.extendModules` instead, and it seems to work. I think this is a hack, so any idea why this happens?

danielphan2003:
Also remove https://github.com/Pacman99/digga/blob/2f6bcd38f148a8ffc367400de532f070a0a74294/flake.nix#L51 since we're not depending on nixos-generators for customBuilds.

blaggacao:
bors try

blaggacao:
bors try

montchr:
#443

blaggacao:
bors try

examples/devos/shell/devos.nix
```diff
@@ -44,6 +44,7 @@ in
     # (docs python3Packages.grip) too many deps
     (docs mdbook)
     (devos inputs.deploy.packages.${pkgs.system}.deploy-rs)
+    (devos inputs.nixos-generators.${pkgs.system}.defaultPackage)
```
blaggacao, line 47:
```suggestion
    (devos inputs.nixos-generators.defaultPackage.${pkgs.system})
```

blaggacao:
bors try

Pacman99:
This should be ready to merge with NixOS/nixpkgs#166148 in which should fix @danielphan2003's error

Users will need to update their own nixpkgs input for the feature.

PR #438, opened 2022-02-28
Add Support for Darwin/macOS Hosts

### LINKED ISSUE #440: nix-darwin support
Ability to configure `darwinConfigurations` through nix-darwin just like nixos hosts.

## Additional context
Discussion in https://github.com/divnix/digga-old/issues/16 (can't transfer that issue here, since it used to be in this repo)

cc @montchr

montchr:
Redo of https://github.com/divnix/digga-old/pull/111

Fixes #440 (used to be divnix/digga-old#16)

## TODO

- [x] Checks succeed in each example flake
	- [x] `hmOnly`
	- [x] `groupByConfig`
	- [x] `devos`
- [x] Fix `Darwinia.tests = [ allProfiles ]` in `groupByConfig` example
- [x] Update `devos` example for compatibility
- [x] Test on a real Apple computer to manage a real system configuration

### Miscellaneous/Cleanup

- [x] Determine next steps for tracking nix-darwin to address missing `lib` option (see https://github.com/divnix/digga-old/pull/111#issuecomment-1014661785)
	- If https://github.com/LnL7/nix-darwin/pull/429 is merged before this PR is ready to go, then update the input URL to point back to the upstream repo
	- If the nix-darwin PR mentioned above is not yet merged, then consider forking nix-darwin in the divnix org
	- Otherwise, continue tracking https://github.com/montchr/nix-darwin/tree/add-toplevel-option-lib and make that a protected branch
- [x] Clean up commit history
- [x] Run `nix build .#mkApiReference*`
- [x] Revert changes made for testing purposes
- [ ] Point the examples' `digga` inputs to release

## Notes

- I wanted to take as much "magic" as possible out of the new changes to `fup-adapter.nix` to make it easier to debug. So there may be some opportunity for introducing some abstractions in that file down the line, but for now I'd like to keep things less complex through repetition.


src/mkFlake/fup-adapter.nix
```diff
@@ -51,20 +51,71 @@ let
...
     inherit (config)
       channelsConfig
       supportedSystems;
     inherit self inputs sharedOverlays;
 
-    hosts = builtins.mapAttrs (_: stripHost) config.nixos.hosts;
+    hosts = builtins.mapAttrs (_: stripHost) (
+      flake-utils-plus.lib.mergeAny
+        nixosHosts
+        darwinHosts
```
montchr, line 117: Resurfacing a suggestion from @Pacman99 in https://github.com/divnix/digga-old/pull/111#discussion_r796270941:

> While I realize that this makes the code easier, the issue with this is that darwin and nixos hosts can't have the same name even though this is actually possible. This creates an artificial restriction on host names. If we do decide to do this, there needs to be an error triggered early when a user passes a host to both `nixos.hosts` and `darwin.hosts` so they know that only one will be created.

And my response:

> Good point, though I assumed that would be an edge case. There is a unique-name restriction amongst `nixos.hosts` already, and I would think that allowing identical hostnames across `{nixos,darwin}.hosts` would be difficult to manage and could lead to weird issues down the road. So I'm thinking that throwing an error for any identical hostnames would be the better option here.

src/mkFlake/fup-adapter.nix
```diff
@@ -75,14 +126,10 @@ let
       )
       config.channels;
 
-    hostDefaults = flake-utils-plus.lib.mergeAny (stripHost config.nixos.hostDefaults) {
```
montchr, line 78: https://github.com/divnix/digga-old/pull/111#discussion_r796272327

src/mkFlake/fup-adapter.nix
```diff
@@ -75,14 +126,10 @@ let
...
-      # add `self` & `inputs` as specialargs so their libs can be used in imports
-      specialArgs = config.nixos.importables // { inherit self inputs; };
-      modules = config.nixos.hostDefaults.exportedModules ++ defaultHostModules;
-    };
```
montchr, line 82: https://github.com/divnix/digga-old/pull/111#discussion_r796276072

shell.nix
```diff
@@ -52,13 +55,13 @@ let
...
         # ensure: restore input
-        [ -z $diggaurl ] || sed -i "s|$tempdigga|$diggaurl|g" flake.nix
+        [ -z $diggaurl ] || ${pkgs.gnused}/bin/sed -i "s|$tempdigga|$diggaurl|g" flake.nix
       }
 
       digga_fixture() {
         # ensure: replace input
         diggaurl=$({ grep -o '"github:divnix/digga.*"' flake.nix || true; })
-        [ -z $diggaurl ] || sed -i "s|$diggaurl|$tempdigga|g" flake.nix
+        [ -z $diggaurl ] || ${pkgs.gnused}/bin/sed -i "s|$diggaurl|$tempdigga|g" flake.nix
```
Pacman99, line 64: Feel free to commit this bit directly to main

montchr:
I'm now using `digga.lib.mkFlake` in all my system configurations, including three darwin systems, one of which is my primary machine. The switch from flake-utils-plus was actually really seamless! More deletions than additions in the diff, which I usually take as a good sign. https://github.com/montchr/dotfield/commit/aa49be37f1404c5fd6886fd988d36e1759feac0b

I'm not yet making use of the home-manager profiles because currently my "user" profiles are laced with a lot of system-scoped config, among other oddities. It's a work in progress.

So with that, I've checked off the "Test on a real Apple computer to manage a real system configuration" task. There's a few more to go but it feels like this is getting pretty close to done 😸 

Next up, I'll be taking a look into the DevOS example and identifying what might need updating there.

Pacman99:
Wow thats amazing :) This is indeed looking close to being done!

You should add your flake to the "In the Wild" section in the README

montchr:
bors try

montchr:
bors try

montchr:
I made some updates in e6286e7b386e1c83d6b93b1fd51730d924c3824c to get the devos example a bit closer, but am currently getting the following:

```
error: a 'x86_64-linux' with features {} is required to build '/nix/store/ds55lzn494dcjsp5lja4dxi6kl5nzkqi-builder.pl.drv', but I am a 'x86_64-darwin' with features {benchmark, big-parallel, nixos-test}
exit 1: command [/nix/store/2gbwlvyk9z8x99xvhn42fnfzif1n42jx-nix-2.5pre20211206_d1aaa7e/bin/nix flake check "$@"] failed
```

Not sure what's causing that but my next step will be trying to use a more-up-to-date nix binary…

Also, @Pacman99 do you have any suggestions as to how I can get bors to run builds prior to an approving code review? As you can see above and in my other open PR, I keep running into that "Already running a review" message.

montchr:
bors delegate+

montchr:
bors try

montchr:
bors cancel

Pacman99:
> error: a 'x86_64-linux' with features {} is required to build '/nix/store/ds55lzn494dcjsp5lja4dxi6kl5nzkqi-builder.pl.drv', but I am a 'x86_64-darwin' with features {benchmark, big-parallel, nixos-test}
exit 1: command [/nix/store/2gbwlvyk9z8x99xvhn42fnfzif1n42jx-nix-2.5pre20211206_d1aaa7e/bin/nix flake check "$@"] failed

This seems like the error that happens from using IFD. I don't see any IFD usage here though. Anything that forces you to build a derivation for a system that your on is an issue.

montchr:
> This seems like the error that happens from using IFD. I don't see any IFD usage here though. Anything that forces you to build a derivation for a system that your on is an issue.

@Pacman99 Yeah this definitely has me a bit confuddled. I've suspected it either has something to do with the devshell or the ISO builder, but at the moment I don't have much to go on. Unfortunately even running `nix flake check --show-trace` directly in `examples/devos` gives nothing more specific than the error you saw. Any ideas how to get a more-helpful backtrace for this error?

montchr:
Also, noting for myself later: CI is using an older version of nix unstable which I've had trouble with on darwin in the past. Make sure it's using something more recent.

```
/nix/store/2gbwlvyk9z8x99xvhn42fnfzif1n42jx-nix-2.5pre20211206_d1aaa7e
```

src/mkFlake/fup-adapter.nix
montchr: Added check in 9ce5e01678387d5849142863bc029ddf4b4cafc4

montchr:
Builds are green! 🎈

examples/devos/flake.nix
```diff
@@ -63,18 +68,30 @@
 
         channelsConfig = { allowUnfree = true; };
 
-        channels = {
-          nixos = {
-            imports = [ (digga.lib.importOverlays ./overlays) ];
-            overlays = [
+        channels =
+          let
+            # TODO: any reason to avoid applying these to `latest` via `sharedOverlays`?
```
Pacman99, line 73: No I don't think so, we just haven't dealt with multiple channels being used at once.

Pacman99 (approved):
It all looks pretty good.

I wonder if the devos core profile could be condensed into one file with an `ifDarwin` and `ifNixOS` flag behind every statement. I'm good with the separating them too like what you've done.

examples/devos/flake.nix
```diff
@@ -24,7 +26,8 @@
       home.url = "github:nix-community/home-manager/release-21.11";
       home.inputs.nixpkgs.follows = "nixos";
 
-      darwin.url = "github:LnL7/nix-darwin";
+      # TODO: update url once https://github.com/LnL7/nix-darwin/pull/429 is merged
+      darwin.url = "github:montchr/nix-darwin/add-toplevel-option-lib";
```
montchr, line 30: @Pacman99 This still hasn't been merged to nix-darwin. What do you think about forking nix-darwin to the divnix org to provide a stable home for this branch?

Alternatively, if possible, it might be preferable to declare the missing lib option within the digga repo until the upstream PR is merged.

Pacman99: I would prefer the latter option. I think it should be pretty safe to declare the lib option even when we're evaluating for NixOS, since the module system can merge options. I think we could declare the lib option with `mkOptionDefault`

bjw-s:
Hi! It may be my incredible n00bness with Nix/digga, but I am not able to get this branch working on my M1 Macbook Pro. I have copied the `examples/devos` folder, and I am not able to get `nix flake check` to work. The error message I get is this:

`error: a 'x86_64-linux' with features {} is required to build '/nix/store/ds55lzn494dcjsp5lja4dxi6kl5nzkqi-builder.pl.drv', but I am a 'aarch64-darwin' with features {benchmark, big-parallel, nixos-test}`

Do you perhaps have a clue as to what I am doing wrong?

Pacman99:
@montchr might be able to give you better insight. But that error usually happens when your using IFD which causes nix to build derivations for other systems while evaluating.

montchr:
@bjw-s do you have a publicly-available example configuration I could take a look at?

Unfortunately, I don't currently have access to an M1 Mac, but I may be able to borrow one to test sometime next week. I've seen a very similar error on an Intel Mac and in GitHub Actions but GitHub Actions doesn't currently provide a runner for `aarch64-darwin`. 

One other thing to note is that everything in the `home` configuration will need to be buildable on darwin. I've worked around some of those errors with conditional checks for `pkgs.stdenv.buildPlatform.isDarwin`.

montchr:
As a side note, I expect to have a good chunk of time this weekend to get this over the line, so I'm hoping to merge this Saturday or Sunday after taking another pass to respond to code review feedback, cleanup, and documenting any caveats and important nuances.

Pacman99:
Awesome! One change I want to make before merging is to have all the templates refer to the digga release instead of main, so we can test main a bit more before people start using it.

Along with that I was wondering how common it would be for someone to use both nixos and Darwin. It might be a good idea to make a separate Darwin specific template instead of editing devos. And groupByConfig could be left as it is demonstrating the possibility of using both together.

diogox:
> Along with that I was wondering how common it would be for someone to use both nixos and Darwin.

I didn't read the rest of the discussion, but I happened to read this in a notification, so I figured I'd drop my 2 cents and say that I'm the exact situation of migrating from NixOS to MacOS because of work, and it would be nice to have the same config work on both, in case I ever manage to go back to NixOS. (Or if I want to have 2 machines, one for each OS)

Hope it helps ✌️

montchr:
I hope we continue to accommodate both NixOS and Darwin hosts in DevOS or build out another example similar to DevOS with a goal of demonstrating cross-compatible profiles and graceful fallbacks for unsupported systems. The Digga examples have always aimed to demonstrate shareable modules/profiles/configs/etc. across hosts. Because of that, it's seemed to me that Digga has always had the potential to provide a guide towards cross-host-type configuration sharing, which was my original reasoning for taking this task on. 

My experience is similar (in reverse) to @diogox's story: I've primarily been a Mac user in the past, and I still need to use Mac for work, but NixOS is where I want to live, as I've become more and more frustrated with the directions taken by Apple/macOS in the past few years and have been feeling the strain of living in Apple World. I wouldn't be surprised if the draw of using Nix/nix-darwin/home-manager on macOS has encouraged others to want to make the same journey, and I'm certain that there are lots of other Linux users out there who are expected to use a Mac for work who would still benefit from a solid example of profile sharing.

src/collectors.nix
```diff
@@ -0,0 +1,29 @@
+{ lib }:
+let
+  collectHosts = nixosHosts: darwinHosts:
```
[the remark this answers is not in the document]
montchr: I was on the fence about where these should go. But I do think they can be useful as exported functions – I'm using `collectHosts` in my own configuration for `homeConfigurations`:

https://github.com/montchr/dotfield/blob/2461ed81bfb3a523b585f9ed9a81d8cda8bc4885/flake.nix#L304-L306

examples/devos/flake.nix
```diff
@@ -42,6 +45,8 @@
       nixos-hardware.url = "github:nixos/nixos-hardware";
 
       nixos-generators.url = "github:nix-community/nixos-generators";
+
+      nixpkgs.follows = "nixos";
```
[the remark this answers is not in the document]
montchr: I don't think so, in fact it's probably outright wrong. This looks like it should accomplish the same goal while taking the host's tracked channel into account:

https://github.com/divnix/digga/blob/2558087132e03faec4e13fda3547e72f55d77332/examples/devos/modules/nix-path.nix#L3

examples/devos/flake.nix
montchr: Updated and fixed – no more fork. I wasn't sure how to leverage `mkOptionDefault` but open to suggestions.

bjw-s:
> @bjw-s do you have a publicly-available example configuration I could take a look at?
> 
> Unfortunately, I don't currently have access to an M1 Mac, but I may be able to borrow one to test sometime next week. I've seen a very similar error on an Intel Mac and in GitHub Actions but GitHub Actions doesn't currently provide a runner for `aarch64-darwin`.

@montchr  Well to be fair, I haven't done anything other (yet) than check out this branch and try to get it to build. On the current revision things seem to work a lot better. My steps to reproduce:
```console
nix-shell https://github.com/divnix/digga/archive/refs/heads/darwin-support.zip
mkdir /tmp/digga && cd /tmp/digga
nix flake init -t github:divnix/digga/darwin-support
nix flake check  # this works on the current revision
nix flake show   # this also works on the current revision
nix build .#darwinConfigurations.Mac.system
```

The last step returns:
```
error: The option `environment.shellAliases.nrb' has conflicting definition values:
       - In `/nix/store/3qc2jan7bjahmiv9h3p84r1ymkkcla8q-source/profiles/core/common.nix': "sudo darwin-rebuild"
       - In `/nix/store/3qc2jan7bjahmiv9h3p84r1ymkkcla8q-source/profiles/core/darwin.nix': "sudo darwin-rebuild switch --flake"
```

How to continue from here? I understand I would need to flesh out the hosts/homes/users/etc, but for testing/learning purposes I figured I could just try to build the provided example `Mac` configuration?

> Along with that I was wondering how common it would be for someone to use both nixos and Darwin.

For me it's similar: I run a Mac as my personal machine, but (currently) run Arch Linux as my work machine (thinking about switching over to nixOS). Having the same way of configuring and potentially sharing configuration snippets between both
machines would be the dream.

montchr:
@bjw-s thanks for the info. I can confirm `nix build .#darwinConfigurations.Mac.system` does not work in `examples/devos`. I'll take a look into this today.

Our CI workflows currently don't `nix build` any of the hosts, which is how this slipped through. I'm not sure why that is exactly, but it seems like an oversight. I'm going to add build steps to the workflow configuration to see how that goes…

montchr:
> Our CI workflows currently don't nix build any of the hosts, which is how this slipped through. I'm not sure why that is exactly, but it seems like an oversight. I'm going to add build steps to the workflow configuration to see how that goes…

On second thought, let's come back to this later… if we want to add builds for the example hosts, it'd be better to rework the CI configuration thoughtfully so as not to repeat checks unnecessarily. Something like filtering workflow runs by changes to the files in `examples/**`.

I've opened #457 to address this later.

montchr:
@bjw-s I've force-pushed some updates to this branch, which includes a fix for the issue you ran into while building the example host.

montchr:
@Pacman99 I think this is ready to merge, aside from rebuilding my nix-darwin system with the latest from this branch, in addition to addressing your suggestion:

> One change I want to make before merging is to have all the templates refer to the digga release instead of main, so we can test main a bit more before people start using it.

What did you have in mind exactly? Since the examples have all been updated for compatibility with this branch, pinning them to the previous version would break. Also feel free to push commits to this branch or open a PR against it. I don't plan on making any more force-pushes at this point so let's consider the history stable.

bjw-s:
> @bjw-s I've force-pushed some updates to this branch, which includes a fix for the issue you ran into while building the example host.

Thanks for the update! I can confirm that I am able to get `nix build .#darwinConfigurations.delphi.system` to work now (`delphi` is my hostname), however, when running `./result/sw/bin/darwin-rebuild switch --flake .` I get this:

```
building the system configuration...
error: The option `_module.args.pkgs' is defined multiple times.

       Definition values:
       - In `/nix/store/2asxcb3djbp24yz7kg2j7lj5m0yqfh3b-source/eval-config.nix'
       - In `<unknown-file>'
```

My repo is [here](https://github.com/bjw-s/digga-test), should that provide any additional insight.

grimsleydl:
If it helps narrow it down, I get the same ```error: The option `_module.args.pkgs' is defined multiple times``` error on an Intel Mac, both with the trivial default config as-is and with the hostname/username adjusted to reflect the actual host.

kid:
that error comes from gytis-ivaskevicius/flake-utils-plus#113

I've been facing the same issue for days while using `flake-utils-plus` directly with `nix-darwin` and only found the culprit today

Pacman99:
Yeah I'm good with merging this directly. I guess I was thinking of a way to encourage people to use the release tag for digga, instead of main. But its fine, the checks pass in this branch and the code all looks good so it should be stable.

montchr:
@bjw-s 

> Additionally, if I set the machine `system` to `aarch64-darwin`, the build will fail because `nix-index` cannot be compiled for it (it seems to be coming from the `core.common` profile).
>

Looks like this is fixed in https://github.com/bennofs/nix-index/commit/0634333254e12d3848f4ee7a7896cae57d5f4cf6 which is not available in the version of `nix-index` pulled from `nixos-21.11`. The stable channel version is [from 2018](https://github.com/bennofs/nix-index/releases/tag/v0.1.2)! Looks like `nixos-unstable` currently has a version that should work. I've added it to `overrides.nix` in e6b14d39d6d2e0a0544dbc50c7cc235acb2bc123 – does that fix the issue you ran into with that package?

> I've found [this gist](https://gist.github.com/jmatsushita/5c50ef14b4b96cb24ae5268dab613050#file-flake-nix-L34) (through this [NixOS forum post](https://discourse.nixos.org/t/simple-workable-config-for-m1-macbook-pro-monterey-12-0-1-with-nix-flakes-nix-darwin-and-home-manager/16834)) where the author is overriding the package arch, but unfortunately I have no idea how I would port this to the devos structure.

As an overlay, it could be added directly to `channels.nixpkgs-darwin-stable.overlays`:

https://github.com/divnix/digga/blob/e6b14d39d6d2e0a0544dbc50c7cc235acb2bc123/examples/devos/flake.nix#L81

Or you could, for example, create a file in the `overlays/` directory, which would be imported automatically by the existing `importOverlays` invocation in `channels.nixpkgs-darwin-stable.imports`.

I don't have a sense of how common a scenario like the one you ran into with `nix-index` might be – at this point, the M1 Macs have been out for a while so I'm guessing a lot of folks have already updated their packages? `nix-index` seems like it might be an outlier since I take it you didn't run into that error with any other packages included by default in `examples/devos/`, and also as mentioned above, the version in the stable channel is very old. Perhaps we could add an overlay like that as a future enchantment later, but I'm not even able to test it at the moment 🙂

montchr:
> If it helps narrow it down, I get the same error: The option `_module.args.pkgs' is defined multiple times error on an Intel Mac, both with the trivial default config as-is and with the hostname/username adjusted to reflect the actual host.

@bjw-s @grimsleydl @kid Assuming that the issue is coming from https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/113 do either of these fixes help?

- https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/113#issuecomment-1068621947
- https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/117

Pacman99:
Thank you for the amazing work @montchr!! Glad to see Darwin support become a reality.

montchr:
Thanks for the support! It's been a journey.

PR #447, opened 2022-04-02
Use nixos-* not release-*

### LINKED ISSUE #399: Updating/Rebuilding occasionally rebuilds way too many packages
I don't fully understand the update flow in devos, so not marking this as a bug as it's probably me misusing the system.

I'm confused about what is causing ~500 packages to be built during updates as I'm using nixpkgs release 21.05 channel for my base system and have picked only few packages from latest channel.

I suspect that the default `haskellPackages` override might be the cause of this overcompilation, by picking plenty of Haskell packages from latest channel. This came with the upstream configuration though, as I hadn't added it myself.

Here is my devos fork: https://github.com/cipharius/devos

Pacman99:
fixes #399


flake.nix
```diff
@@ -7,7 +7,7 @@
 
   inputs =
     {
-      nixpkgs.url = "github:nixos/nixpkgs/release-21.11";
+      nixpkgs.url = "github:nixos/nixpkgs/nixos-21.11";
```
montchr, line 10: Considering folks (including myself) are often confused by the various branches, maybe we add a link and/or brief explanation behind the decision to track the `nixos-*` branches?

Reading through #399, it looks like https://nixos.org/manual/nixpkgs/stable/#overview-of-nixpkgs provides a decent explanation.

montchr (approved):
🧙

flake.nix
Pacman99: good idea, I added this comment to all files:
`# Track channels with commits tested and built by hydra`

PR #449, opened 2022-04-09
devos: only format modified lines in git pre-commit hook

teutat3s:
In my git committing workflow, I strive to make atomic commits, sometimes staging only single lines in a file.

The current git pre-commit hook is not aware of this, it just stages the whole modified `*.nix` file, even if it was only partly staged before.

This PR is an attempt to fix this and only format lines that were modified by leveraging `git stash`.

Feedback welcome!


montchr (commented):
I've been trying to verify the described behavior but am still seeing the entire file being reformatted upon commit. Are you able to put together a demo or provide some testing steps?

teutat3s:
> I've been trying to verify the described behavior but am still seeing the entire file being reformatted upon commit even after only staging part of the file with magit or `git add -p`. Are you able to put together a demo or provide some testing steps?

Did you test this with a `*.nix` file? If so, and it doesn't work for you, I'll try to put together a test case :+1:

montchr:
Yeah, still running into the old behavior.

This is, for context, with the current version of `examples/devos` copied to a separate location and turned into its own git repo since the hooks won't work otherwise (that's a whole nother issue...)

But actually, I'm wondering if the shell script syntax fixed in #480 is causing a problem here too... I'll try testing both together.

montchr:
@teutat3s I just tested this again, still no luck. I put together an asciicast to demo the process.

Rough summary:

```console
# setup
cp -R digga/examples/devos ./devos-testing
cd devos-testing
git init # to set up hooks
git add . && git commit -m 'init'

# verification of correct hook version from this PR
# asciicast only shows this for a second but you can pause
bat $DEVSHELL_GIT_HOOKS_DIR/bin/pre-commit

# add a bunch of dumb formatting in three separate locations
vim flake.nix

# stage only one hunk of the three
git add -p

# verify index contents
git diff --staged

git commit -v -m 'testing'

# shows that all hunks have been committed
# only one of the hunks should have been committed
git status
```

[asciicast](https://asciinema.org/a/utX6coT9UKxdYrQNdV98q16na)

examples/devos/shell/hooks/pre-commit.sh
```diff
@@ -5,21 +5,32 @@ then
...
+  new_stash=$(git rev-parse --quiet --verify refs/stash)
+
+  # Format staged changes
   nixpkgs-fmt "${nix_files[@]}" \
   && git add "${nix_files[@]}"
+
+  # If unstaged changes were stashed re-apply to working tree
+  if [ "$old_stash" != "$new_stash" ]; then
+      git stash pop --quiet
+  fi
```
montchr, line 30: Since this is a pre-commit hook, these stashed changes might be added back to the index prior to the commit? Does this need to move to a post-commit hook?

montchr (changes requested):

teutat3s:
AFAICT, this has been fixed by https://github.com/divnix/digga/pull/491

PR #444, opened 2022-03-27
Revert devshell subflakification

montchr:
Recent changes to `main` moved the devshell to a separate flake in a subdirectory. This caused numerous issues, initially on `main` but especially on #438 and Darwin hosts. It also broke `nix develop` unexpectedly by requiring the user specify the new path to the devshell flake as `./devshell#default`, which was not documented.

### Context

- #443 
- Discussion in 5b33c4baefe44379b80592fbf531304ff3b513b7
- Conversation in Matrix channel [[link](https://matrix.to/#/!luwEFvTJxfOhBlnMbr:myrdd.info/$4HW3Q7GMYibs2YOQre5_o0I5muQ1MqNrHxQl6I0Hf2w?via=matrix.org&via=pub.solar&via=myrdd.info)]

### Details

This PR reverts the following commits:

- 5b33c4baefe44379b80592fbf531304ff3b513b7
- 85496bd4deb2392c2abb27f889c0a090967ad8ff
- 94cc51e2b795780f5b44f89e9c3f06d5debf898a
- 75c0f266f46e0a3d110cec5b146c0ee48393cd2f
- 9da3840a6f4c86c3058f8bd2e1d824dbee2abdfe

PR #453, opened 2022-04-21
Format with alejandra

Pacman99:
alejandra seems like the best choice out there and our codebase would look a lot cleaner with this.

This should probably be merged after #438, so we don't have to deal with the conflicts.

The first commit is me, the second commit is all alejandra.

nix 2.8 isn't in nixpkgs yet, but you can always do `nix develop -c fmt`


devShell.nix
```diff
@@ -30,120 +29,117 @@ let
...
+      alejandra
+      nixUnstable
+    ];
+
+    env = [
+      {
+        name = "NIX_CONFIG";
+        value = ''          extra-experimental-features = nix-command flakes
+                  extra-substituters = https://nrdxp.cachix.org https://nix-community.cachix.org
+                  extra-trusted-public-keys = nrdxp.cachix.org-1:Fc5PSqY2Jm1TrWfm88l6cvGWwz3s93c6IOifQWnhNW4= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs='';
```
montchr, line 102:
```suggestion
        value = ''
          extra-experimental-features = nix-command flakes
          extra-substituters = https://nrdxp.cachix.org https://nix-community.cachix.org
          extra-trusted-public-keys = nrdxp.cachix.org-1:Fc5PSqY2Jm1TrWfm88l6cvGWwz3s93c6IOifQWnhNW4= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs=
        '';
```

devShell.nix
```diff
@@ -30,120 +29,117 @@ let
...
+      _PATH=''${PATH}
+      # Load installed profiles
+      for file in "$DEVSHELL_DIR/etc/profile.d/"*.sh; do
+        # If that folder doesn't exist, bash loves to return the whole glob
+        [[ -f "$file" ]] && source "$file"
+      done
+      # Exert exorbitant privilige and leave no trace
+      export PATH=''${_PATH}
+      unset _PATH
+    '');
```
montchr, line 119: I've been eyeing this for a bit: is it still necessary? Looks like https://github.com/numtide/devshell/pull/123 is fixed.

montchr:
I haven't yet used https://github.com/numtide/treefmt, but it looks pretty useful as "One CLI to format the code tree". @blaggacao originally introduced its usage in the devshell changes reverted in #444. Looks like it could handle formatting for `*.nix` files with `alejandra`, in addition to other formatters per file type.

Considering formatting changes are usually pretty disruptive, it might be worth resurrecting the usage of `treefmt` in this PR since the `treefmt.toml` file there also added formatter configuration for `*.sh` files with `shfmt`, and we have a few of those in the repo. Plus, Prettier for all the rest.

https://github.com/divnix/digga/pull/444/files#diff-83489924125e52fef286509dc792070b570a0c656cd9b048445f87bebc57405a

nrdxp:
Since treefmt was mentioned, we have a pretty useful solution at work that uses a devshell pre-commit hook to call treefmt on the entire repo before any commits are made. I don't think anybody would mind if we still it:
https://github.com/input-output-hk/devshell-capsules/blob/main/pre-commit.sh

I know we had something very similar here before, not sure if we still do

Pacman99:
Definitely, I'll setup treefmt here.

We still have the pre-commit hook for devos, but it was never added to digga's devshell, I can do that here to. And we can switch to treefmt in the hook too. Right now it uses nixpkgs-fmt.

devShell.nix
Pacman99: Yeah I'm not entirely sure about this either. I'll remove it and check if anything breaks.

blaggacao: Yes, this is fixed. Thx!

Lord-Valen: Removed in #490.

Lord-Valen:
This can be closed since #491 was merged.

PR #454, opened 2022-04-22
devShell: use --override-input to set digga url

Pacman99:
This method is a lot simpler and improves devShell readability


montchr:
nice! works for me locally (except for #475).

once i've merged #476 i'll rebase and force push to get builds green.

montchr (approved):

PR #459, opened 2022-04-24
458 Onboarding docs get you to a nix-shell

### LINKED ISSUE #458: Onboarding documentation incomplete
New here and a bit confused. Have been reading the nix & flake documentation for a couple days. I'm trying to create a new pc setup based on digga (or devos?) since this seems like the best way to have a nice flake setup.

## Expected behavior
Following the getting started docs should get you a working repo that makes use of digga

## Current behavior
I'm following `doc/start` on my NixOS machine, so according to `doc/start/index.md` I run
```
nix flake init -t github:divnix/digga
```
and then add the result to a git repository (based on my prior reading I know this is so that nix will recognize the new flake)
```
git init
git add .
git commit -m init
```
But then the next line in the docs is perplexing:
> 
**This will place you in a new folder named devos with git initialized, and a nix-shell that provides all the dependencies, including the unstable nix version required.**

So it seems this line is from the old devos documentation, and the instructions above it are new, for installing digga (as of #448).

If I run `nix-shell` I get an error, so probably some additional step is needed?
```
error: 'outputs' at /nix/store/...-source/flake.nix:48:5 called without required argument 'digga'
```

If I ignore this and follow Next Steps to `doc/start/from-nixos.md` the first line there is `bud up`, but I don't think I can use `bud` quite yet. (Is bud part of digga, or just devos? The `flake.nix` in the digga template I pulled includes a `bud.url`, so I'm guessing it's a digga thing, but what is the difference between digga and devos?)

## Possible solution
Document what is expected to happen after downloading the template.

I'm going to try figuring out how to fix the `nix-shell` error above and then try to run `nixos-rebuild switch --flake .` on it?

## Context
#442 #448 

@Pacman99

pgarrison:
Fixes #458

The command for users without flakes support is untested, I just modified the earlier `flake init` command using the digga shell

### DIFF
--- a/doc/start/index.md
+++ b/doc/start/index.md
@@ -19,9 +19,18 @@ git add .
 git commit -m init
 ```
 
-This will place you in a new folder named `devos` with git initialized, and a
-nix-shell that provides all the dependencies, including the unstable nix
-version required.
+To drop into a nix-shell, if you don't have flakes setup, use the digga shell to create a `flake.lock`:
+```sh
+nix-shell "https://github.com/divnix/digga/archive/main.tar.gz" \
+  --run "nix flake lock"
+```
+Or if you do have flakes support, just run:
+```sh
+nix flake lock
+```
+
+Finally, run `nix-shell` to get to an interactive shell with all the dependencies, including the unstable nix
+version required. You can run `menu` to confirm that you are using digga (expected output includes [docs], [general commands], [linter], etc.).
 
 In addition, the [binary cache](../integrations/cachix.md) is added for faster deployment.
 


Pacman99:
Ahh I see the problem, good catch. Once you clone devos, you should just be able to run `nix-shell` in the folder, so you don't need to use the digga shell then.

rwaltr:
Thank you for the PR. Been stumbling with this until I found this.

rwaltr:
Thank you for the PR. Been stumbling with this until I found this.

pgarrison:
@Pacman99 sorry I'm new here, can you clarify the difference between cloning devos vs cloning digga?

Pacman99:
Cloning devos is when you pull the template, cloning digga isn't needed unless you plan on contributing.
Either way I was just thinking out loud. This is definitely an important fix. Thanks for the PR!

PR #460, opened 2022-04-28
devos/darwin: refine users

montchr:
I meant to include these changes in #438:

- Demonstrates the ability to define reusable home-manager user configurations
  via `home.users` and import them into any arbitrary home-manager user within a
  `nixos` or `darwin` profile.
- Rename the `darwin` user to `primary` to distinguish it from all the other
  references to 'darwin' and also to aid in demonstrating the flexibility of
  "portable" home-manager users
- Use Digga's `collectHosts` function to... collect hosts for deriving `homeConfigurations`
- Remove some extraneous comments


Pacman99:
Could we remove the nixos user and use admin/primary everywhere? It would make writing docs easier too, since we only have to reference one user to change.

montchr:
yeah, good point. will update.

montchr:
> Could we remove the nixos user and use admin/primary everywhere?

I took a halfway approach, keeping the `nixos` user around but only using it in the `bootstrap-iso` host since that's the expected user for live media.

montchr:
Now I'm remembering that one of the reasons I split up the users was because nix-darwin lacks a lot of the basic `users.users.<name>.*` options. `extraGroups`, for one. Builds are failing during checks for that reason.

https://github.com/divnix/digga/runs/7312053064?check_suite_focus=true#step:7:312

PR #465, opened 2022-05-16
Fix Darwin

diogox:
[This PR] was just merged on `flake-utils-plus` to fix an issue with the darwin integration.

This PR simply updates that dependency to make use of it.


Pacman99:
Thanks!

PR #468, opened 2022-05-21
docs: minor formatting updates for function and option descriptions

montchr:
also removes the unnecessary `userType` variable

### DIFF
--- a/doc/api-reference-channels.md
+++ b/doc/api-reference-channels.md
@@ -7,7 +7,7 @@ Configure your channels that you can use throughout your configurations.
 
 
 ## channels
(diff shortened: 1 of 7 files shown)


montchr:
I came back to this recently, wondering why I had left it in a draft state. It looks like these changes actually have an overall detrimental effect on the docs’ readability due to the way the various variables are composed to form an interpolated doc string. For example, capitalizing the first word in an “options” variable often results in weird mixed capitalization in the resulting options. So this is not the way to go (an acceptable “solution” would require backtracking a bunch on this branch).

PR #472, opened 2022-06-26
Update home-manager to master

GTrunSec:
solved https://github.com/nix-community/home-manager/blob/master/flake.nix#L31

Notice: not ready to merge in digga.

### DIFF
--- a/examples/devos/flake.nix
+++ b/examples/devos/flake.nix
@@ -10,6 +10,7 @@
       # Track channels with commits tested and built by hydra
       nixos.url = "github:nixos/nixpkgs/nixos-22.05";
       latest.url = "github:nixos/nixpkgs/nixos-unstable";
+
       # For darwin hosts: it can be helpful to track this darwin-specific stable
       # channel equivalent to the `nixos-*` channels for NixOS. For one, these
       # channels are more likely to provide cached binaries for darwin systems.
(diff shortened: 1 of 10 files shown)


montchr:
While this has been bugging me too, and I need it fixed for my own use cases, I'm not sure whether `digga@main` and its examples should officially support `home-manager@master` until 22.11 because doing so now would also require tracking `nixpkgs-unstable`. IIRC, we've done that in the past and it's been... unstable.

Technically, I cheat and track `home-manager@master`+`nixos-22.05` as `nixpkgs`, but that's an unsupported configuration. Usually it's fine, but it's bit me in the past so I maintain my own personal fork for weird workarounds.

`nixpkgs-unstable` aside, this would also be a breaking change to users whose projects aren't tracking unstable/master. So it seems best to wait, IMO.

But considering all the above... maybe we should continue maintaining a release/develop/trunk branch to have a semi-official fix for this issue and others like it? I know we've been down that road too, but this kind of thing seems like a good reason to have a more visible experimental branch for others to refer to. Thoughts? cc @Pacman99 @blaggacao

montchr:
@GTrunSec could you revert 5fce9b8 and reformat with `nix develop --command -- fmt` please?

GTrunSec:
sure, thanks for your work.

montchr:
@GTrunSec thank _you_ for putting this together. I opened a PR in your fork adding some suggestions.

https://github.com/GTrunSec/digga/pull/1

@a12l I would recommend setting `home.stateVersion = "22.05";` in your configuration if you haven't already. The state version setting is used by home-manager to help retain backwards compatibility.

PR #476, opened 2022-07-09
support nixos 22.05 + remove bud and misleading onboarding docs (via #477)

montchr:
22.05 is the current version of NixOS, so we should update digga to match.

fixes #475

PR #477, opened 2022-07-09
remove bud and misleading onboarding docs

### LINKED ISSUE #418: Flakes check errors on unsupported system
I'm currently evaluating devos as my base template for my systems. I can `nix build` all my systems now, but `nix flake check` gives an error.

## Expected Behavior

`nix flake check` and everything that needs it, including `deploy` should work.

## Current Behavior

Error on checking:
```
error: Package ‘wlrobs-unstable-2021-05-13’ in /nix/store/626qldjl59n3cx00l2sd86734vdj99pk-source/pkgs/applications/video/obs-studio/plugins/wlrobs.nix:20 is not supported on ‘i686-linux’, refusing to evaluate.

       a) To temporarily allow packages that are unsupported for this system, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowUnsupportedSystem = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowUnsupportedSystem = true; }
       to ~/.config/nixpkgs/config.nix.
```
Interestingly enough, sometimes (e.g. after running `bud update`) we get a different error:
```
error:
       Failed assertions:
       - Must use Linux for modules that require systemd: emacs, foot, gammastep
       - The module wayland.windowManager.sway does not support your platform. It only supports

         - aarch64-linux
         - armv5tel-linux
         - armv6l-linux
         - armv7a-linux
         - armv7l-linux
         - i686-linux
         - m68k-linux
         - mipsel-linux
         - powerpc64-linux
         - powerpc64le-linux
         - riscv32-linux
         - riscv64-linux
         - s390-linux
         - s390x-linux
         - x86_64-linux
```
## Steps to Reproduce

1. Checkout the branch https://github.com/pimeys/nixos/tree/devos
2. Run `nix flake check`

## Your Environment
See the posted repo (devos branch) for details.

montchr:
This has been planned for some time.

It seems necessary to do so now because after updating inputs to track
NixOS 22.05 channels and releases, checks fail unexpectedly, with the
traces leading back to `bud`. Disabling its modules allows checks to
run successfully.

Rather than figuring out what's going wrong with `bud`, it seems more
appropriate to remove it as planned.

Because this is closely related to updating Digga/Devos to support 22.05,
I've set #476 as a base branch.

Many many people have complained about the misleading documentation.
Rather than continuing the path of "we need to fix this", we can reduce
the harmful effects of the inaccurate docs by removing them all
together.

Since much of the confusion in the "getting started" docs came from
`bud`-specific instructions, and `bud` is being removed, removing the
inaccurate docs also seems like it can be a kick towards replacing them.

Also:

- The NixOS manual is the canonical guide to installation. We don't need
to repeat its instructions.
- The numerous references to networking configuration don't have a place
within Digga -- I'm assuming these are `bud`-specific things which are
no longer relevant with `bud` gone.

fixes #418


nrdxp (approved):
LGTM. @blaggacao correct me if I'm wrong, but you probably don't have time maintain bud atm, so removing it is probably the right move at this point.

PR #487, opened 2022-10-15
Add another example repo to README

sweenu:
Hello!

I propose adding my repo as another example repo in the README.
The interesting stuff would be the fact that I mix my home-manager configs with my system configs and maybe the RaspberryPi deployment.

I won't feel bad if you reject the PR ;)


Pacman99:
Looks great, I've been meaning to setup authelia, it's nice to see an example

PR #491, opened 2022-11-07
:brain: `redesign` Use treefmt

Lord-Valen:
Similar to #453 but using treefmt. Formatting should be applied with an additional commit or PR, probably the latter (it becomes hard to read the changes otherwise).


Pacman99 (approved):
Lgtm, thanks for getting this to happen! Does this PR include formattting digga library code?

Lord-Valen:
To reduce diff clutter, this PR doesn't include any changes made by formatters (they also make merge conflicts hell). We can run fmt later.

blaggacao (approved):
LGTM, for the actual formatting commit, do: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

flake.nix
```diff
@@ -148,6 +147,7 @@
       '';
 
       # digga-local use
+      formatter = nixlib.lib.genAttrs supportedSystems (s: nixpkgs.legacyPackages.${s}.alejandra);
```
Lord-Valen: @Pacman99 What is the purpose of this?

blaggacao: This gets alejanda for all supported systems onto `nix fmt` jobs

shell/default.nix
```diff
@@ -0,0 +1,118 @@
...
+
+  test = type: name:
+    withCategory "tests" {
+      name = "check-${name}";
+      help = "Checks ${name} ${type}";
+      command = ''
+        set -e
+        # set -x
+
+        tempdigga=path:$PRJ_ROOT
```
Lord-Valen, line 44: https://github.com/divnix/digga/actions/runs/3448153423/jobs/5754901537#step:7:37

The tests are apparently failing because tempdigga is not stored in the nix store?

Lord-Valen: Could it be the extra backmatter in PRJ_ROOT, as seen [here](https://github.com/divnix/digga/actions/runs/3448153423/jobs/5754901537#step:7:11), causing failures unique to remote tests? Locally, PRJ_ROOT is much simpler, as `/home/lord-valen/dev/digga` and the tests do not fail in this manner.

Lord-Valen:
A lock file was added, causing the test failures. Strange that `.gitignore` did not catch it.

Lord-Valen:
> A lock file was added, causing the test failures. Strange that `.gitignore` did not catch it.

Annoyingly, .gitignore files in subdirectories override the .gitignore files in parent directories.

blaggacao:
It's done! Further fixups will have to raise the PR count.⏩🤘

Thank you!

blaggacao:
Since you seem.to be interested in good developer tooling, there's also `conform` & `lefthook` as adopted in `divnix/std`...

PR #493, opened 2022-12-08
Update home-manager to 22.11

jokogr:
This commit updates the home-manager flake input to the 22.11 release and creates additional support for the new homeManagerConfiguration setup that is introduced there.

~The new release of home-manager introduces a dependency to numtide/flake-utils, which already exists in deploy-rs. To reduce the amount of different versions, this commit makes the input of deploy-rs follow the respective one of home-manager.~ (see my comment below on why I dropped this)

Regarding homeManagerConfiguration, some arguments that were previously in use, now need to be used inside the new modules argument (check [here][hm-22.11-highlights] for more details).

The stateVersion is additionally explicitly defined to be the same as the home-manager's release version, so that existing configurations do not break when they upgrade their home-manager flake input.

[hm-22.11-highlights]: https://nix-community.github.io/home-manager/release-notes.html#sec-release-22.11-highlights

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,8 +15,9 @@
     deploy.url = "github:serokell/deploy-rs";
     deploy.inputs.nixpkgs.follows = "nixpkgs";
 
-    home-manager.url = "github:nix-community/home-manager/release-22.05";
+    home-manager.url = "github:nix-community/home-manager/release-22.11";
     home-manager.inputs.nixpkgs.follows = "nixlib";
+    home-manager.inputs.utils.follows = "flake-utils";
 
     darwin.url = "github:LnL7/nix-darwin";
     darwin.inputs.nixpkgs.follows = "nixpkgs";
--- a/src/mkFlake/outputs-builder.nix
+++ b/src/mkFlake/outputs-builder.nix
@@ -27,20 +27,27 @@
     homeDirectory = "${homeDirectoryPrefix}/${username}";
   in
     home-manager.lib.homeManagerConfiguration {
-      inherit username homeDirectory pkgs system;
+      inherit pkgs;
 
-      extraModules = config.home.modules ++ config.home.exportedModules;
+      modules =
+        config.home.modules
+        ++ config.home.exportedModules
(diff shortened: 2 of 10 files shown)


flake.nix
```diff
@@ -14,8 +14,9 @@
 
     deploy.url = "github:serokell/deploy-rs";
     deploy.inputs.nixpkgs.follows = "nixpkgs";
+    deploy.inputs.utils.follows = "home-manager/utils";
```
jokogr, line 17: NixOS/nix#6549 suggests (and I have also just tested) that this would not work if a configuration repository which uses digga, overrides digga's inputs, so I am going to drop this and have two versions of numtide/flake-utils in flake.lock.

src/mkFlake/outputs-builder.nix
```diff
@@ -25,23 +25,53 @@
...
+          config.home.modules
+          ++ config.home.exportedModules
+          ++ [
+            ({
+                imports = [configuration];
+
+                home = {
+                  inherit username homeDirectory;
+
+                  stateVersion = lib.mkDefault homeManagerReleaseVersion;
```
Sohalt, line 62: This will change the `stateVersion` when users update the home-manager release and don't set `stateVersion` explilcitly.
I think we should not set `stateVersion` at all and have users do that as part of their configuration.

Lord-Valen:
The current CI failure is not a result of changes made in this PR. Rebasing now or after #497 should resolve the issue.

blaggacao:
#497 is merged.

jokogr:
I have rebased this PR and took @Sohalt's suggestion to keep the stateVersion open to the users. In the meantime, I have also simplified the home-manager configuration in Digga by keeping only the 22.11 version and updated the home-manager input in every example.

jokogr:
> I have also simplified the home-manager configuration in Digga by keeping only the 22.11 version

Checks are passing, but let me know if you prefer the previous version, so that Digga users could still use home-manager 22.05 and Digga from the `main` branch.

PR #499, opened 2023-01-25
:pencil: devos: clarify how to use /lib/

ProducerMatt:
Hi all. I wanted to put custom functions in the /lib/ directory, but lib/default.nix was basically empty and I couldn't find anything in the docs about it. After a while I figured it out, and I thought it would help future noobs if there was a note in the file explaining itself.

Thanks for the great template!


Pacman99:
Good point, thanks!

